PR #101, opened 2018-12-06
added the key in lang files

NikhitaPaturu:
added the wordpress key for events and groups lang files


resources/lang/en/group-audits.php
```diff
@@ -36,6 +37,7 @@
             'latitude' => '<strong>Latitude</strong> was modified from "<strong>:old</strong>" to "<strong>:new</strong>"',
             'longitude'   => '<strong>Longitude</strong> was modified from "<strong>:old</strong>" to "<strong>:new</strong>"',
             'free_text' => '<strong>Free Text</strong> was modified from "<strong>:old</strong>" to "<strong>:new</strong>"',
+            'wordpress_post_id' => '<strong>Wordpress post ID</strong> was modified to "<strong>:new</strong>"',
```
ngm: The old value should be displayed here as well

Issue #108, opened 2019-03-04
Group associations aren't removed when an account is deleted


ngm:
This causes an error on /group/all, when it tries to access the name of the soft-deleted user.  (Probably elsewhere too.)

We should: 

- [ ] remove group associations when an account is deleted
- [ ] have proper null checks wherever we're currently just assuming the user will always exist

PR #117, opened 2019-05-29
Resolution for #108

spt1224:
Add getFullName() on the UserGroups model to fix Volunteers list when user is soft-deleted.
Remove group associations when soft deleting a user.


app/UserGroups.php
```diff
@@ -93,4 +93,13 @@ public function scopeConfirmedInvitation($query)
     {
         return $query->where('status', 1)->orWhereNull('status');
     }
+
+    public function getFullName()
```
ngm: Is this called anywhere?

spt1224: Yes, but perhaps not anymore. These calls were the cause of the error that appeared when a user was soft deleted:

https://github.com/TheRestartProject/restarters.net/blob/master/resources/views/partials/volunteer-row.blade.php#L30
https://github.com/TheRestartProject/restarters.net/blob/master/resources/views/partials/volunteer-badge.blade.php#L27

ngm: Ah OK, bit confused as from diff it looks as if it's a completely new method.

spt1224:
Well it is a completely new method, but it was being called. Maybe
$volunteer used to be a User instance rather than UserGroup?

Issue #112, opened 2019-05-14
Notification when account has been deleted


ngm:
* When a user deletes their account, admins should receive a notification that this has happened
* Admins should be able to opt-in to receive these notifications or not
* The Notification to be sent via email and database channels
* Notification to contain name of user

ngm:
This has been done and merged.

Issue #111, opened 2019-05-14
Optional Admin notification when photos are uploaded to an event


ngm:
* A new Notification that is sent when photos are added to an event
* Notification only sent to Admins who have opted-in to receive this notification type, 
* The Notification action should take the admin to the edit page for that event, so as to see the photos for the event

PR #118, opened 2019-06-04
Event Photo Upload Notifications (#111) and User Deletion Notifications (#112)

spt1224:
Unit testing to follow tomorrow once I've researched how to properly unit test triggered events and notifications.

This builds on issue #108 (PR #117), so should finalize that PR first.


spt1224:
Test cases have been pushed.

ngm:
@spt1224 Thanks Sean, the notification is being sent fine.

However, when the user is deleted, the notification shows 'Deleted User' as the user's name.

User details are anonymised when the user is deleted - so it looks as if the email is sent after this anonymisation takes place.  In the notification we need to see the name of the user before it has been anonymised.

ngm:
@spt1224 For the in-app notification, the deleted user notification has the name of the deleted user hyperlinked, but the link doesn't go anywhere.  Can the link be removed just for this notification?

ngm:
Image upload notification works well - I like the extra checks for not notifying the user if they uploaded the image themselves, and throttling to stop multiple notifications being sent!

PR #163, opened 2020-05-01
Updates MiscCat status queries, status view and tests

zenlan:
Completes DAT-252

Add status info for devices with "Misc" consensus/majority opinions


zenlan:
Cancel this one, I found a bug

PR #170, opened 2020-07-29
Implements handling for Mobifix completion

zenlan:
Redirects to mobifix/status with message when there are no more records to fetch
Implements default status message when Mobifix task is complete
Hides 3-2-1 opinions counter when Mobifix task is complete

### DIFF
--- a/app/Console/Commands/SyncFaultTypes.php
+++ b/app/Console/Commands/SyncFaultTypes.php
@@ -13,7 +13,7 @@ class SyncFaultTypes extends Command
      * @var string
      */
     protected $signature = 'faultcat:sync {category : [computers|mobiles]}';
-    
+
     /**
      * The console command description.
      *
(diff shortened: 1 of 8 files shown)


app/Console/Commands/SyncFaultTypes.php
```diff
@@ -38,31 +38,43 @@ public function __construct()
...
+                $result = $Faultcat->updateDevicesWithEmptyProblem();
+                if ($result) {
+                    $this->info($result . ' rows updated with Unknown');
+                } else {
+                    $this->info('0 rows updated with Unknown');
                 }
                 break;
             case 'mobiles':
                 $this->info('Updating devices with mobile fault types');
                 $Mobifix = new Mobifix;
```
ngm, line 61: Just to note I got an error here - adding use App\Mobifix to using statements resolved it.

PR #173, opened 2020-08-11
DOT-813 Event count past

edwh:
The count of events wasn't correct for events in the past.  Refactor the method to use Eloquent and fix the problem along the way.


edwh:
Thanks.  Will remove the duff tests.

You're right to pick me up on the other two things, which I'll look at 
tomorrow.  I think the Laravel ways to handle them (but correct me if 
I'm wrong):
For event_timestamp, set up a dynamic field on Party so that it gets 
constructed on the fly from the two separate fields.
For the devices, the idiomatic way is to have a dynamic property using 
relationships, so that the extra data is only fetched when required.  I 
checked that's already set up for devices, but it definitely needs some 
testing, so I should probably build a test script that involves at least 
one of the caller codes.  The $devices parameter can then be removed.

edwh:
I've added an accessor for the event_timestamp, and tested that plus also that the devices are fetched on demand as we'd expect.

That's a reasonable UT of this change.  But how should we test this more widely?  I need a bit of advice there.

ngm:
> That's a reasonable UT of this change. But how should we test this more widely? I need a bit of advice there.

For this case, as I'm a bit unsure as to where this method is used, I would suggest you do a search through the codebase for wherever Party::ofThisGroup is being used, and listing those call locations here.  

Then, if you're able to run the app and navigate to the corresponding pages where those calls are ultimately being used, firstly just to smoke test that they aren't crashing or looking odd.  

And I'll also have at the listed locations and see if anything in my memory gives extra context to them.

edwh:
This is used in the following places:

- loginRegisterStats.  Tested by going to the site logged out.
- AdminController.  Tested from /admin.  Looks ok to me, though it doesn't have the uploaded images so it looks a bit messed up.  The live site allows me to access this page (should it?), and is also missing images.  You can see the number of parties thrown has dropped - do we need to worry about hours volunteered also including the future?
- HostController.  I can't find a route that references this.  I may be forgetting something about Laravel routing, or perhaps this is dead code?
- OutboundController.  That throws an apparently unrelated error when I visit it, including on the live site.

It would be useful for my account on the live site to have suitable permissions so that I can compare

edwh:
I also worried about the timezone safety of event_timestamp.  Are we assuming that all times are viewed by someone in the same timezone as the event?

ngm:
> AdminController.  Tested from /admin.  Looks ok to me, though it doesn't have the uploaded images so it looks a bit messed up.  The live site allows me to access this page (should it?), and is also missing images

Wow! /admin is completely dead code.  That page hasn't been in use since 2018 :D 
So the AdminController::index method and corresponding route can be removed.

> HostController. I can't find a route that references this. I may be forgetting something about Laravel routing, or perhaps this is dead code?

Yes this is also dead code and can be removed.

> OutboundController. That throws an apparently unrelated error when I visit it, including on the live site.

Looking at the corresponding view, I'm certain OutboundController::index is dead code too and can be removed.  info is also used, and visualisationEmbed appears to do nothing so could be removed too.

ngm:
> I also worried about the timezone safety of event_timestamp. Are we assuming that all times are viewed by someone in the same timezone as the event?

Great point - we currently do have that legacy assumption unfortunately, but we have an upcoming epic to introduce proper timezone support across the application.

edwh:
Ok - I've pruned the dead code and done some merging (which pulls in some minor other changes).  Can you check it again now, please?

PR #174, opened 2020-08-12
DOT-1256 Language switcher

edwh:
Simple CSS fix to position language switcher.  Tested on Dashboard page only.

I'm not sure where you put CSS generally, but there's already some for the language-bar, so I've put it near that.


edwh:
Sorry, that's not intentional.  Will fix that.

edwh:
I'm sure there's a better way of doing this but I'm going to close this pull request and open a new one with an updated branch.

PR #175, opened 2020-08-12
DOT-1256 Position language bar correctly

edwh:
Simple CSS fix to position language switcher. Tested on Dashboard page only.

I'm not sure where you put CSS generally, but there's already some for the language-bar, so I've put it near that.


edwh:
That's showing an odd comment change, but it looks the same.

edwh:
Ok.  Do you check in the built CSS too or does that happen as part of 
deployment?

ngm:
> Ok. Do you check in the built CSS too or does that happen as part of deployment?

Yes, check in the built CSS too please.  Not currently built as part of deployment.

edwh:
Ok - back to you for review.

PR #178, opened 2020-08-14
DOT-812 Allow host to be added to event later

edwh:
I tested this by:

- Creating an event in the past
- Removing myself from the list of attendees
- Trying to re-add myself; I don't show in the list

The code looks as though it was intentionally excluding the logged-in user, but I don't think that's correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ restarters-anonymised.sql
 /.idea/restarters.net.iml
 /.idea/vcs.xml
 /.idea/webServers.xml
+/.idea/sshConfigs.xml
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -1064,17 +1064,16 @@ class PartyController extends Controller
     public function getGroupEmails($event_id, $object = false)
     {
         $group_user_ids = UserGroups::where('group', Party::find($event_id)->group)
-        ->where('user', '!=', Auth::user()->id)
-        ->pluck('user')
-        ->toArray();
+            ->pluck('user')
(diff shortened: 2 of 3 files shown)


edwh:
Why should you not be able to invite yourself, if you have rights to do 
so but are not currently associated with the event?

ngm:
> Why should you not be able to invite yourself, if you have rights to do so but are not currently associated with the event?

Good question - in that case, you would generally just click 'RSVP' when viewing the event in the application - inviting yourself and then clicking RSVP in the invitation email would be superfluous if your intention is to just RSVP to the event.

That said - it could actually be of use to invite yourself as part of a mass invite you are making, as a sanity check that the invite has gone out correctly, you didn't make any errors you need to correct, etc.  Though the chances of you not being already associated to the event anyway (given you have probably created in) is quite low. I guess either way, this doesn't cause any trouble.  So hmm, thinking about it, I'm tempted to leave it as it is for now.  What do you think?

edwh:
I could be wrong, but I think you would normally have been associated 
with the event when you created it, wouldn't you?

In which case the only way you would show up on the invitation list is 
if you intentionally removed yourself.  In that case I think it's fine 
for you to appear on the invitation list, as you're deliberately doing 
something a bit odd - for example setting up an even which someone else 
is going to host for you and then changing your mind.

ngm:
> I could be wrong, but I think you would normally have been associated with the event when you created it, wouldn't you?

Precisely, yep.  

> as you're deliberately doing something a bit odd

I concur - right we'll keep this as is then, thanks for the discussion!

PR #183, opened 2020-09-02
Forgot password colour

edwh:
We've been talking as though this was device/browser specific, and that's why we needed Browserstack.  But I think it's a general problem with the alert colours, which reproduces on Chrome.

_alert.scss defines a font colour of #FFF and different background colours for the alert variants. The background colour for alert-success would be unreadable wherever it was used.  Given that alert-primary has a background colour of $brand, I've set that for alert-success too.  That gives this:

image

That's a different change than James suggested, but I think it's probably better to stick to the brand colours.

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -16428,9 +16428,10 @@ table.sortable > thead th div.mozilla {
 }
 
 .alert-success {
+  color: #222;
+  border-radius: 0px;
   background-color: #fff4e4;
   border: 3px dashed #222;
-  border-radius: 15px;
   padding: 20px 45px 20px 35px;
   font-size: 16px;
   overflow: hidden;
--- a/resources/assets/sass/_alert.scss
+++ b/resources/assets/sass/_alert.scss
@@ -43,7 +43,7 @@
     }
 
     &-success {
-        background-color: $brand;
+        background-color: #fff4e4;
         border: 3px dashed #222;
         border-radius: 15px;
         padding: 20px 45px 20px 35px;
--- a/resources/assets/sass/_alert.scss
+++ b/resources/assets/sass/_alert.scss
@@ -115,7 +115,7 @@
 }
 
 .information-alert {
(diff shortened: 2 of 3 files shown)


ngm:
Although it resolves the alert box styling, including the global css on header_plain seems to introduce some new problems, e.g. the layout of /user/recover changes from:

image

to

image

and /user/register changes from

image

to

image

So it seems like there a number of styles that the global css is including that we don't want to pull in for the plain header layout.  Can we fix the alert without including the global css here?

edwh:
Ok. I've pushed a simpler fix - see what you think.

This is classic tech debt stuff - the multiple CSS files and the logged 
in/out differences are confusing.

public/css/app.css
```diff
@@ -16428,9 +16428,10 @@ table.sortable > thead th div.mozilla {
 }
 
 .alert-success {
+  color: #222;
```
ngm, line 16431: These changes need to be somewhere in the sass files (perhaps resources/assets/sass/_alert.scss?) - public/css/app.css is built from those files and overwritten each time they're built.

ngm (changes requested):

edwh:
Of course.

Any reason not to delete public/css/app.css from the codebase?  It's 
just confusing clutter.

ngm:
> Any reason not to delete public/css/app.css from the codebase?  It's just confusing clutter.

I think the only reason at present is that it isn't built as part of the current deployment process.  (Where we push to a bare repo on the server, and a post-receive hook copies the files to the application folder).  

So there's no **good** reason... but we can't do it immediately, as we would need to improve the deploy process to include the built css/js files somehow first.

public/css/app.css
edwh: Pushed.  Have a look now?

ngm (approved):
LGTM :+1:

PR #184, opened 2020-09-02
Vue infrastructure

edwh:
This PR makes changes which are necessary to allow us to use Vue.  It includes the first use, which is the event impact stats.

Things to be aware of:

- Normally you'd have one Vue instance, but it's ok to have multiple.  Anything with a class of "vue" will get a Vue instance created on it.
- The translations from the PHP lang/ folder are packaged up during the build and passed to the client, so that they're accessible there.
- It includes bootstrapvue, so that we can use the more idiomatic Vue method of using components from Bootstrap.

The only thing I'm concerned about is that the Vue dev environment seemed to result in extra error checking for undefined variables.  I had to fix a couple of errors in app.js which were blocking it loading.  There's a possibility that this may come back and bite us in other places where we have undefined variables.  So watch out for console errors about undefined variables, and best not to put it live until it's been kicked around a bit.


composer.json
```diff
@@ -12,6 +12,7 @@
         "barryvdh/laravel-translation-manager": "^0.5.1",
         "doctrine/dbal": "^2.7",
         "fideloper/proxy": "^4.0",
+        "fzaninotto/faker": "1.9.1",
```
ngm, line 15: Interesting that this has moved from require-dev, wonder why a faker is needed for production code and not just in tests.

resources/assets/js/app.js
```diff
@@ -1293,42 +1294,6 @@ function initAutocomplete() {
     tokenFieldCheck();
   });
 
-  function deviceFormCollect($form) {
```
ngm, line 1296: Some of these changes seem related to the work in the unpowered branch, and not just the event stats, so I'm hoping this will all merge in OK?

ngm (approved):
All looks good to me, only issue for me is merging resources/assets/js/app.js with the unpowered branch, I get a couple of minor conflicts that I'm not sure how they should be merged.  Let's go through that together.

composer.json
edwh: I had some weird issues where this kept not installing properly.  I decided to be lazy and let webpack tree-shaking handle it.

edwh: Can dig further if you want.

PR #186, opened 2020-09-21
New design for details/attendance on upcoming events

edwh:
WIP - comment to follow.


resources/assets/js/mixins/co2equivalent.js
```diff
@@ -1,21 +1,5 @@
 export default {
   methods: {
-    pluralise(str, val) {
```
ngm, line 3: Are we using a different method for pluralisation now?

ngm: Ah yep, it's a couple of files down :)

resources/assets/js/store/attendance.js
```diff
@@ -0,0 +1,49 @@
+import Vue from 'vue'
+
+const axios = require('axios')
```
ngm, line 3: Is axios preferable to fetch()?

resources/assets/js/store/index.js
```diff
@@ -0,0 +1,11 @@
+import Vue from "vue";
+import Vuex from "vuex";
```
ngm, line 2: Is it Vuex that provides the store?

resources/assets/sass/_events.scss
```diff
@@ -323,29 +323,30 @@
     background-color: $brand-white !important;
 }
 
-#devices-section {
-    border: 1px solid #0394a6 !important;
+.ourtabs {
```
ngm, line 326: Is this ourtabs as in, not bootstrap's tabs?

resources/views/events/view.blade.php
```diff
@@ -1,290 +1,363 @@
...
+          }
+
+          $expanded_hosts = [];
+          foreach ($hosts as $host) {
+            $thisone = $host;
+            $thisone['volunteer'] = $host->volunteer;
+            $expanded_hosts[] = $thisone;
+          }
+
+          $attendance_edit = (FixometerHelper::hasRole(Auth::user(), 'Host') && FixometerHelper::userHasEditPartyPermission($formdata->id, Auth::user()->id)) || FixometerHelper::hasRole(Auth::user(), 'Administrator');
```
ngm, line 178: Name could be clearer, e.g. can_edit_attendance

resources/views/events/view.blade.php
```diff
@@ -1,290 +1,363 @@
...
+              }
+            ?>
+            <div class="vue">
+              <EventImages :images="{{ json_encode($collected_images)}}" />
+            </div>
+          @endif
 
-            <h2>@lang('events.event_details')</h2>
-            <div class="card events-card">
+          @if( $event->isInProgress() || $event->hasFinished() )
```
ngm, line 207: Isn't this from the previous work on event stats?  Just wondering why it's popping up in this pull request.

resources/views/events/view.blade.php
```diff
@@ -433,4 +507,4 @@ class="btn btn-link dropdown-toggle"
   @include('includes.modals.event-add-volunteer')
   @include('includes.modals.event-request-review')
 
-  @endsection
+@endsection
```
ngm, line 510: A lot of the changes displayed in the diff here are whitespace changes, makes it a bit hard to review actual changes.  Wonder if there's an easy way to avoid that - let's have a think at next retro.

ngm (commented):
Code review looks good - just a couple of questions for my own understanding of how some things work - can go through at retro.

Can go to functionality review with James.

resources/assets/js/store/attendance.js
edwh: Yes.  You can do basic API calls with fetch() as you can with $.ajax, but axios lets you add things like retrying, or headers which are sent on all requests, or callbacks which intercept all your requests, or timeouts, or, or, or.

resources/assets/js/store/index.js
edwh: Yes, exactly.

resources/assets/sass/_events.scss
edwh: Our overrides of bootstrap styling.  Would you like a different name to make that clearer?

resources/views/events/view.blade.php
edwh: Agreed, will change.

resources/views/events/view.blade.php
edwh: You're right, this shouldn't be in this arm of code, as it doesn't apply to upcoming events anyway.  Will remove.

resources/views/events/view.blade.php
edwh: I think there may be an ignore button somewhere in github's review for whitespace.  I don't find diff format great for reviewing, as I like to see the context, so I tend to review more complex PRs in my IDE (PhpStorm) where I get a side-by-side diff.  So maybe the answer is a different review tool?

As for this one, I think it's probably my IDE tidying up - the indentation was off in the original.

edwh:
Thanks.  I've pushed those changes.

edwh:
I think this might need a re-review - I've done a bunch of stuff in response to feedback from James, and also some refactoring to make a follow-on change for past events easier.

PR #188, opened 2020-09-25
Attendance past new design

edwh:
This has the new design for past event details and attendance.  The key differences are:

- You can see the counts for numbers of volunteers, and the function that goes along with that such as the warning.
- Various minor design stuff, e.g. handle long names better, shrink "See all attended" etc to just "See all" as there isn't enough room on some screen sizes.
- Show HTML event descriptions, including truncation which shouldn't break if it happens to truncate halfway through an HTML tag.


ngm:
Not sure what happened with this PR, but the code has definitely been merged in and been live for a while.

PR #194, opened 2020-10-05
DOT 1282: unpowered group impact stats

ngm:
Reuses (and renames) the EventStatsValue and EventStatsImpact components.

Added in a prop for statsEntity on StatsImpact to determine whether its for a group or an event, to then subsequently pull in the right translation string.

Added fixed width to stats images.

Started the branch from master, but have done the pull request against dev2.  Have merged dev2 in to this branch (as well as some commits from `group-details-volunteers`).


resources/assets/js/components/GroupStatsFacts.vue
```diff
@@ -0,0 +1,88 @@
...
+  components: {StatsValue},
+  props: {
+    stats: {
+      required: true,
+      type: Object
+    }
+  },
+  computed: {
+    translatedFacts() {
+      return this.$lang.get('groups.group_facts')
```
edwh, line 26: To avoid "1 events", you need to pluralise this - see for example GroupVolunteers.  Please think about whether this applies to other stats in this change too.

app/Group.php
```diff
@@ -251,28 +251,46 @@ public function getGroupStats($emissionRatio)
...
+
+            $fixed_devices += $eventStats['fixed_devices'];
+            $fixed_powered += $eventStats['fixed_powered'];
+            $fixed_unpowered += $eventStats['fixed_unpowered'];
+            $repairable_devices += $eventStats['repairable_devices'];
+            $dead_devices += $eventStats['dead_devices'];
+            $no_weight += $eventStats['no_weight'];
+            $devices_powered += $eventStats['devices_powered'];
+            $devices_unpowered += $eventStats['devices_unpowered'];
         }
```
edwh, line 282: Are there any stats on the event which we do _not_ want to sum for the group?  We could just iterate through $eventStats rather than list them individually.

resources/assets/js/components/EventDescription.vue
```diff
@@ -9,8 +9,6 @@
   </CollapsibleSection>
 </template>
 <script>
-import { DATE_FORMAT } from '../constants'
-import moment from 'moment'
```
edwh, line 13: I discovered yesterday that moment isn't being actively evolved.  At some point we should switch to dayjs, as it's a lot smaller.  That's a tech debt task for much later, if/when we move to a fully VueJS app.  Do we have a place to dump things like that as we go?

resources/assets/js/components/StatsImpact.vue
```diff
@@ -7,29 +7,33 @@
...
   mixins: [ co2equivalent ],
   props: {
     stats: {
       required: true,
       type: Object
+    },
+    statsEntity: {
+      required: true,
+      type: String
     }
```
edwh, line 37: FYI, there isn't (so far as I know) a built-in "Enum" type.  If you're being super-careful, you can add a custom validator to enforce the usage.  I don't think you need to do that here but it's worth skimming the way it's done.

https://vuejs.org/v2/guide/components-props.html#Prop-Validation

resources/lang/en/groups.php
```diff
@@ -98,13 +99,17 @@
...
+  'about_none' => 'There is no description for this group yet.',
+  'read_more' => 'READ MORE&nbsp;<img class="icon" src="/images/add-icon.svg" style="width: 20px;" />',
+  'read_less' => '<img class="icon" src="/images/minus-icon.svg" style="width: 20px;" />&nbsp;READ LESS',
   'volunteers' => 'Volunteers',
   'invite_to_group' => 'Invite to group',
   'key_stats' => 'Key stats',
-  'participants' => 'Participants',
-  'hours_volunteered' => 'Hours volunteered',
-  'total_events' => 'Total events',
+  'group_facts' => 'Group facts',
```
edwh, line 108: It's what the design says, but "facts" is a dull word.  "Achievements"?

edwh (changes requested):
Looks good - most comments are FYIs.  Nice to make the component more generic - I didn't feel I knew the rest of the code well enough to do that yet.

resources/assets/js/components/GroupStatsFacts.vue
ngm: Ah thanks yes, good point - this would apply to 'participants' and 'hours volunteered' stats too.

app/Group.php
ngm: I'll double check them - I don't think there's any event stats we need to exclude from the group stats.  

There's one currently not being included in the event stats, hours volunteered - but would make sense to add this to event stats, and then yes we could iterate through $eventStats.

resources/assets/js/components/EventDescription.vue
ngm: A ticket in Jira in the 'Platform - Infrastructure Issues' epic would be a good place to log tech debt - although I know Laurence is keen not to have tickets created for 'nice to haves', only for things we do actively plan on doing (at some point), because the backlog gets a bit unmanageable otherwise.

At the same time, it's nice not to lose ideas... perhaps we could have a tech debt doc for the nice to haves?  And if something keeps on popping up repeatedly, then it's more than a nice to have?

resources/lang/en/groups.php
ngm: Agreed!  I'll change it to achievements.

resources/assets/js/components/EventDescription.vue
edwh: Yes, a tech debt doc - probably better for us to have the raw stuff somewhere so that we can sort through to turn it into Jira tickets as and when.  Vue refactoring is the key tech debt task for now anyway.

resources/assets/js/components/GroupStatsFacts.vue
ngm: I've changed the translatedTitle computed property to make use of the `choice` method from the Laravel JS Localization package, using the roundedCount property to determine whether it's singular or plural.  Given that count is required this seems OK for now.  Works here and looks fine on the existing stats on the events page too - does it look OK to you?

 https://github.com/rmariuzzo/Laravel-JS-Localization#support-for-singular-and-plural-message-based-on-a-count

app/Group.php
ngm: I've changed this now to iterate through the keys from the event stats.

resources/assets/js/components/GroupStatsFacts.vue
edwh: That sounds better than what I did in the pluralize.js mixin that I added.  Should we switch to your model throughout?

ngm: I think we should - from reading the link above, Lang.js is based on Laravel's Translator class, so all of the existing pluralisation / choice translation templates we have should work as is with Lang.choice.

edwh: Ok, I've pushed a commit for that to this branch.

PR #195, opened 2020-10-13
Convert devices section to use Vue

edwh:
This is a large change.

- Introduce a store for events and devices.
- Rewrite the device section into Vue components.  Two-way binding is a bit of a pain.
- Small changes to API calls to return more useful data.
- Make some properties fillable which weren't previously (and probably couldn't be edited).
- Improve handling of useful URLs so that you can add multiple on the create.
- Ensure photo upload works, which resolves some problems with earlier design changes.
- Introduce EventPage so that we can move more out of the view blade.
- Delete code (all of it?) which is now redundant.


resources/assets/js/app.js
```diff
@@ -1396,178 +1397,6 @@ function initAutocomplete() {
 
     });
 
-    $('.add-device').on('submit', function(e) {
```
ngm, line 1399: Nice to see all this go!

ngm (approved):
Much of the Vue changes are currently beyond my ken (i.e. store related things), you know far more than me on this area, so approving as is given James has signed off on functional testing.

PR #197, opened 2020-10-22
Dot 1346 safari timepicker

ngm:
@edwh would be great to get your feedback on this code!

For DOT-1346, the issue was that we're using inputs with type=time, and that Safari on desktop doesn't support this.

I've used bootstrap-vue's b-form-timepicker component to replace the time range selection.  I've used a hidden input who's value is bound to the value of the timepicker in order to pass the value to the server on form submit.

I've used the agent checker library we have so that we only use this component on desktop Safari, and stick with the browser's own component elsewhere (chiefly because I think it's better to use the browser's native components when they do exist).  Side note: it would be better to do a check on support for input type=time rather than check for a specific browser, but I don't know how to do that.

The other issue I hit was our global CSS overriding lots of bootstrap's style rules and making the components a mess.  I've tried to reset this in the component to make it look sensible again.  Though I've realised I would need to do the same here with the date picker (also not supported on Safari desktop), so if there's a way to *not* inherit any of the parent application's CSS inside a component that would be great - I think you'd mentioned something similar once but it might have been different.

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -1724,7 +1724,8 @@ jQuery(document).ready(function () {
         'groupdescription':  require('./components/GroupDescription.vue'),
         'groupvolunteers':  require('./components/GroupVolunteers.vue'),
         'groupstats': require('./components/GroupStats.vue'),
-        'eventtimerangepicker': require('./components/EventTimeRangePicker.vue')
+        'eventtimerangepicker': require('./components/EventTimeRangePicker.vue'),
+        'eventdatepicker': require('./components/EventDatePicker.vue')
       }
     })
   })
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -1723,7 +1723,8 @@ jQuery(document).ready(function () {
         'groupheading':  require('./components/GroupHeading.vue'),
         'groupdescription':  require('./components/GroupDescription.vue'),
         'groupvolunteers':  require('./components/GroupVolunteers.vue'),
-        'groupstats': require('./components/GroupStats.vue')
+        'groupstats': require('./components/GroupStats.vue'),
+        'eventtimerangepicker': require('./components/EventTimeRangePicker.vue')
       }
     })
   })
--- a/resources/assets/js/components/EventDatePicker.vue
+++ b/resources/assets/js/components/EventDatePicker.vue
@@ -1,0 +1,48 @@
+<template>
+    <div>
+        <b-form-datepicker class="datepicker" v-model="value"></b-form-datepicker>
+        <input type="hidden" name="event_date" :value="value" />
+    </div>
+</template>
+
+<script>
+export default {
+  props: {
+    'initialvalue' : {
+      required: false,
+      type: String
+    }
+  },
+  data() {
+    return {
+      value: this.initialvalue,
+    }
+  },
+}
+</script>
+
+<style scoped lang="scss">
+@import 'resources/global/css/_variables';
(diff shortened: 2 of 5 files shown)


edwh:
Surprising that Safari doesn't support this.

It might be possible to test for this on the client size; see https://modernizr.com/docs/#inputtypes . I've not used modernizr much but it could be worth a shot if you want to make it a little less ugly.

I generally avoid _id_ values in components, so that the components can be used multiple times within a single page.  I think I've seen a few examples where Bootstrap advises use of an id value for accessibility, but that doesn't seem to be the case here.  ids tend to be used a lot in cases where you have global JS, but they aren't usually necessary in components.

Increasingly I am avoiding Bootstrap column layouts in favour of flex/grid unless I'm dealing with the level of the whole page layout.  You could do this, for example:

```
  <div class="d-flex">
    <b-form-timepicker id="start-time" v-model="startTime" placeholder="--:--" @input="changeEndTime" />
    <input type="hidden" name="start" :value="startTime" />
    <b-form-timepicker id="end-time" v-model="endTime" placeholder="--:--" />
    <input type="hidden" name="end" :value="endTime" />
  </div>

```

Global CSS overriding styles in components is a pain, and I've hit it elsewhere.  There isn't a better solution than you have, i.e. override them back again.  I'm thinking that if and when we get enough stuff moved over into Vue then we could have a radical rationalisation and prune of the global CSS, but it's too soon for that at the moment.  

On the function:
* Do you want to guard against people setting the end time before the start time?
* Any minimum duration/time steps?
* The existing input has a colour when you focus - do you want to preserve that?
* If you wanted to be fancy, you could handle the case where someone changes the start time and set the end time to the corresponding time based on the duration before the change, if you see what I mean.

ngm:
Thanks @edwh - I've updated the component to make use of classes rather than ids internally, and to use flex rather than Bootstrap's rows/cols.

> On the function:

    Do you want to guard against people setting the end time before the start time?
    Any minimum duration/time steps?
    The existing input has a colour when you focus - do you want to preserve that?
    If you wanted to be fancy, you could handle the case where someone changes the start time and set the end time to the corresponding time based on the duration before the change, if you see what I mean.

All of those logic changes would be very sensible changes - but given that right now this is just to avoid bugs for Safari specifically, I'll leave it with the existing behaviour.  Perhaps if we switch this entire page to Vue at some point, we can introduce improved logic then (whether we reuse this component for all browsers or continue to use native inputs).

modernizr looks really handy if we find ourselves needing to check this kind of thing a lot - will stick with agent for now though, as it is all set up already.

edwh:
All fine.

I've been using ::v-deep rather than /deep/, though I think they are 
identical.

ngm:
> All fine. I've been using ::v-deep rather than /deep/, though I think they are identical.

I saw it mentioned actually that /deep/ is being deprecated, however when I tried ::v-deep it didn't seem to work - let me try again...

ngm:
> I saw it mentioned actually that /deep/ is being deprecated, however when I tried ::v-deep it didn't seem to work - let me try again...

@edwh Yep the styles not being picked up for me when using ::v-deep rather than /deep/  - are they definitely working for you?  Perhaps I have something missing in my npm build steps?  

I came across this unanswered question on the same problem on SO  - https://stackoverflow.com/questions/60693777/or-v-deep-not-working-in-vue-project

edwh:
That's interesting.  I've had some v-deep problems.  Next time I have 
one, I'll try /deep/ instead.

PR #198, opened 2020-10-26
DOT-1309 Group events new design

edwh:
Rework the group events section (and standalone page) into the new design.


resources/assets/js/components/GroupEvents.vue
```diff
@@ -0,0 +1,291 @@
...
+            <p v-if="!upcoming.length">
+              {{ translatedNoUpcoming }}.
+            </p>
+            <b-table-simple v-else sticky-header="50vh" responsive class="pl-0 pl-md-3 pr-0 pr-md-3 pb-2 mb-2" table-class="m-0 leave-tables-alone">
+              <GroupEventsTableHeading />
+              <b-tbody class="table-height">
+                <GroupEventSummary v-for="e in upcomingToShow" :key="'event-' + e.idevents" :idevents="e.idevents" :canedit="canedit" />
+              </b-tbody>
+            </b-table-simple>
+            <div class="text-center" v-if="limit">
```
ngm, line 38: I guess `limit` relates to my previous question about why we removed the `take` from the backend query.

resources/views/events/index.blade.php
```diff
@@ -20,27 +20,6 @@
           </div>
       @endif
 
-      <div class="row mb-30">
-          <div class="col-12 col-md-12">
-              <div class="d-flex align-items-center">
-                  <h1 class="mb-0 mr-30">
-                      @lang('events.events')
```
ngm, line 27: This seems to have lost the title, doodle, create event button, from the Events page  (/party) - presumably unintentional?

resources/views/events/index.blade.php
```diff
@@ -71,48 +50,49 @@
         @endif
         {{-- END Events to Moderate (Admin Only) --}}
 
-        {{-- Upcoming events for your Groups --}}
```
ngm, line 74: Likewise we have lost Upcoming events from /party ?

resources/views/events/index.blade.php
```diff
@@ -143,42 +123,6 @@
         </section>
         @endif
 
-
-        {{-- Past events --}}
-        <section class="table-section" id="events-4">
```
ngm, line 148: And past events from /party

ngm (changes requested):
Mostly just a few questions for my own interest of Vue workings.

However, functionally, we seem to have lost some elements from the main /party page - I'm presuming this is not an intentional change.

resources/assets/js/components/CalendarAddModal.vue
```diff
@@ -0,0 +1,112 @@
...
+    </b-alert>
+    <div class="d-flex justify-content-between flex-wrap mt-4 mb-4">
+      <b-btn variant="link" href="https://talk.restarters.net/t/fixometer-how-to-add-repair-events-to-your-calendar-application/1770">
+        {{ translatedFindOutMore}}
+      </b-btn>
+      <b-btn variant="link" :href="editUrl">
+        {{ translatedSeeAll }}
+      </b-btn>
+    </div>
+    <template slot="modal-footer" slot-scope="{ cancel }">
```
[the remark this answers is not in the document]
edwh: It's a way of passing things from the component into the slot so that they can then be accessed from within the slot.  The bootstrap modal component comes with its own cancel() method which closes the modal; here we pass it in to the slot and then invoke it on click.

resources/assets/js/components/CalendarAddModal.vue
```diff
@@ -0,0 +1,112 @@
...
+      this.copied = true
+    },
+    copyFail() {
+      this.failed = true
+    }
+  }
+}
+</script>
+<style scoped lang="scss">
+@import 'resources/global/css/_variables';
```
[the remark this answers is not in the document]
edwh: None.  I'll remove.

resources/assets/js/components/GroupEvents.vue
```diff
@@ -0,0 +1,291 @@
...
+              <div class="d-flex justify-content-between">
+                <div>
+                  <b>{{ translatedUpcoming }}</b> ({{ upcoming.length }})
+                </div>
+              </div>
+            </template>
+            <p v-if="!upcoming.length">
+              {{ translatedNoUpcoming }}.
+            </p>
+            <b-table-simple v-else sticky-header="50vh" responsive class="pl-0 pl-md-3 pr-0 pr-md-3 pb-2 mb-2" table-class="m-0 leave-tables-alone">
```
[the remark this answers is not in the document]
edwh: Yes, exactly.  There's also v-else-if.  So here we are putting a message if there's nothing to show, otherwise we build a table.

resources/views/events/index.blade.php
edwh: Yes, you're entirely right.

resources/views/events/index.blade.php
edwh: I missed this function to show all events from your groups.  Have added.

app/Http/Controllers/GroupController.php
```diff
@@ -539,12 +539,10 @@ public function view($groupid)
         //Event tabs
         $upcoming_events = Party::upcomingEvents()
         ->where('events.group', $group->idgroups)
-        ->take(5)
```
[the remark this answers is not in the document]
edwh: Generally I've removed the server-side limiting/pagination.  Sometimes this is because we now show all the events on the client; sometimes it's because we need at least the count of the events.  

We could handle this more efficiently, but it seems overkill to be trying too hard in that area because the number of events we're dealing with at any one time is fairly low (at most a few hundred).

If you're worried about server load and want me try harder in this area, just let me know.

PR #202, opened 2020-10-28
DOT 870 Translate device categories

edwh:
This translates device categories and clusters, by adding extra strings into the existing lang JSON file.  If categories change then we'll need to add them here as well as in the DB; if that happens a lot then we could have a separate script that's run as part of the build.

We have previously discussed having a cron job to recompile the client overnight to pick up new translations.  That would apply to this too.


ngm (changes requested):
Translations working on:

- /fixometer
- /category

Not yet working on:

- e.g. /device/page-edit/18270 (where you are taken if you click on a device row on /fixometer
- e.g. /party/view/1527 (adding/editing items for an event)
- e.g. /group/view/1 (device breakdown section - although maybe this comes after 1284)

edwh:
I've merged in 1284 so I can do that one, and done some more.  Can you 
have another look?

edwh:
Wrong branch - closing.

PR #204, opened 2020-11-09
DOT 870 translate device categories

edwh:
This translates device categories and clusters, by adding extra strings into the existing lang JSON file. If categories change then we'll need to add them here as well as in the DB; if that happens a lot then we could have a separate script that's run as part of the build.

We have previously discussed having a cron job to recompile the client overnight to pick up new translations. That would apply to this too.


resources/assets/js/app.js
```diff
@@ -48,6 +48,7 @@ import lang from 'lang.js';
 import translations from './translations.js';
 const Lang = new lang()
 Lang.setFallback('en')
+console.log("Translations", translations)
```
ngm, line 51: Do we want to leave this in?

resources/assets/js/components/DeviceCategorySelect.vue
```diff
@@ -79,15 +79,17 @@ export default {
         cluster.categories.forEach((c) => {
           if ((this.powered && c.powered) || (!this.powered && !c.powered)) {
             categories.push({
-              name: c.name,
+              // Need to translate.
```
ngm, line 82: Is this a TODO and can be removed now that we are translating?

ngm (changes requested):
Couple of very minor comments, and also noticed that cluster names are not translating on the single group page - 

image

edwh:
Thanks.  Fixed the bottom two.

On the first one, I introduced translation strings into groups.php when I did the redesign for this section.  But you're right, we can now use the same cluster names.  Requires a CSS tweak to make a bit more space, as the earlier translations had "&" rather than "AND".  

Pushed that now.

PR #205, opened 2020-11-09
DOT 1349 bad password help

edwh:
There is an error message already in the DOM, but no code to show it.  Added some.


ngm (changes requested):
This looks like it shows all validation errors even if the only one is related to the passwords, e.g. 
image

edwh:
Try now?

ngm (approved):

PR #206, opened 2020-11-09
Group events page is empty

edwh:
This page got broken when I created GroupPage.vue.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -87,10 +87,9 @@ class PartyController extends Controller
             $upcoming_events_in_area = null;
         } else {
             $upcoming_events = Party::upcomingEvents()->where('users_groups.user', Auth::user()->id)
-                ->take(3)
                 ->get();
 
-            $past_events = Party::UsersPastEvents([auth()->id()])->paginate(10);
+            $past_events = Party::UsersPastEvents([auth()->id()])->get();
 
             if ( ! is_null(Auth::user()->latitude) && ! is_null(Auth::user()->longitude)) {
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/PartyController.php
```diff
@@ -87,10 +87,9 @@ public function index($group_id = null)
             $upcoming_events_in_area = null;
         } else {
             $upcoming_events = Party::upcomingEvents()->where('users_groups.user', Auth::user()->id)
-                ->take(3)
                 ->get();
 
-            $past_events = Party::UsersPastEvents([auth()->id()])->paginate(10);
```
ngm, line 93: Pulling back all results has a bit of a performance hit on the /party page - logged in as a user with 263 past events, it takes 7 seconds to render on my laptop, so will presumably also have the performance issue James logged for his mobile.  Not sure if there's anything we can do here right now?  But we'll need to address this soon I think.

app/Http/Controllers/PartyController.php
```diff
@@ -87,10 +87,9 @@ public function index($group_id = null)
...
+            $past_events = Party::UsersPastEvents([auth()->id()])->get();
```
ngm, line 92: Could we also sort by event_date, most recent first?

ngm (changes requested):
Fixes the issue, but wondering if there's anything performance related we can do regarding the removal of the pagination?

app/Http/Controllers/PartyController.php
ngm: It turns out we are sorting on the query, but I think we lose this when putting items in to the store?  Have put a sort at the point of display for now, but probably a more efficient way of doing this.

edwh: Will discuss at retro.

PR #207, opened 2020-11-10
WIP: microtasking dashboard

ngm:
- might be that we want to have a 'dashboard block' component.  similar to collapsible section
  - using CollapsibleSection at the moment, works OK but with a few differences
    - e.g.here we need backgrounds, borders, etc.
    - e.g. we need a block with a featured image
    - CollapsibleSection has grey line when collapsed on mobile, doesn't really work for these dashboard blocks
- having to include bootstrap breakpoint functions in lots of components?
- passing a lot of properties through page to lower components
  - should I be putting things in a store?
- how do I get the talk url from .env in to the vue components as a constant?
- need to do something about non-logged in layout - it is missing some global styles
- still need to do i18n


ngm:
Closing in favour of #219 which is against dev2 branch.

PR #211, opened 2020-11-16
REP-58 assign editor role

edwh:
This allows Repair Directory SuperAdmins and RegionalAdmins to set the permissions for other Repair Directory users.


app/Policies/UserPolicy.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return mixed
+     */
+    public function canChangeRepairDirRole(User $perpetrator, User $victim, int $role)
+    {
+        # We have rules for whether you can change the Repair Directory role.  Code is structured for readability
+        # of these rules, rather than a single big if.
+        #
+        # Default to forbidden.
+        $ret = FALSE;
+        error_log("Can change role? {$perpetrator->repairdir_role()} . {$victim->id} vs {$perpetrator->id}" );
```
ngm, line 28: If this is to be kept I would use Log::info from Illuminate\Support\Facades\Log (https://laravel.com/docs/5.6/logging)

app/Role.php
```diff
@@ -15,6 +15,11 @@ class Role extends Model
     const GUSET = 5;
     const NETWORK_COORDINATOR = 6;
 
+    const REPAIR_DIRECTORY_NONE = 0;
```
ngm, line 18: Will this set the Repair Directory role to 0 if they are set as having no access?  I think this will cause an error on the Repair Directory Doctrine mapping, as it will try and find a role entity with id of 0.  Whereas if repairdir_role is null, it will ignore it.

app/User.php
```diff
@@ -73,6 +76,11 @@ public function role()
         return $this->hasOne('App\Role', 'idroles', 'role');
     }
 
+    public function repairdir_role() {
+        // Make sure we don't return a null value.  The client select would struggle with null values.
+        return $this->repairdir_role ? $this->repairdir_role : Role::REPAIR_DIRECTORY_NONE;
```
ngm, line 81: Ah looks like it does prefer 0 to null then - we'll have to figure a way around this.  (Either we make Repair Directory handle 0 as a valid role id, or we make the select work with null values.)

app/User.php
```diff
@@ -441,6 +449,26 @@ public function isDripSubscriber()
       return ! is_null($this->drip_subscriber_id);
     }
 
+    public function isRepairDirectoryNone()
+    {
+        return $this->repairdir_role == Role::REPAIR_DIRECTORY_NONE;
```
ngm, line 454: This looks like it is expecting repairdir_role of 0 to be set in the DB?  What would happen if it was null?

app/Policies/UserPolicy.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Determine whether one user can change the Repair Directory role of another to a specific value.
+     *
+     * @param  \App\User  $user
+     * @param  \App\User  $user
+     * @param  int  $role
+     * @return mixed
+     */
+    public function canChangeRepairDirRole(User $perpetrator, User $victim, int $role)
```
ngm, line 21: I think the naming convention would be to have the method called 'changeRepairDirRole' (as when we call in a blade template, we might use @can('changeRepairDirRole', ...)

app/Http/Controllers/UserController.php
```diff
@@ -245,6 +251,38 @@ public function postProfilePasswordEdit(Request $request)
...
+
+        $id = $request->input('id');
+        $role = intval($request->input('role'));
+        $user = User::find($id);
+
+        // Check that we are allowed to change the role, based on our own role.
+        try {
+            $this->authorize('canChangeRepairDirRole', Auth::user(), $user, $role);
+        } catch (\Throwable $e) {
+            error_log($e->getMessage());
```
ngm, line 274: Use Log::error

app/Http/Controllers/UserController.php
```diff
@@ -245,6 +251,38 @@ public function postProfilePasswordEdit(Request $request)
...
+            return redirect()->back()->withErrors($validator)->withInput();
+        }
+
+        $id = $request->input('id');
+        $role = intval($request->input('role'));
+        $user = User::find($id);
+
+        // Check that we are allowed to change the role, based on our own role.
+        try {
+            $this->authorize('canChangeRepairDirRole', Auth::user(), $user, $role);
```
ngm, line 272: I think we might not need this try/catch, as I believe $this->authorize will trigger some Laravel error handling to redirect with a 403 Forbidden if the authorization fails.  

If we just catch the exception and log it, wouldn't we then continue executing the rest of the code and still end up updating the role?

resources/views/user/profile-edit.blade.php
```diff
@@ -33,13 +33,16 @@
...
           <a class="list-group-item list-group-item-action" id="list-account-list" data-toggle="list" href="#list-account" role="tab" aria-controls="account">@lang('profile.account')</a>
 
           <a class="list-group-item list-group-item-action" id="list-email-preferences-list" data-toggle="list" href="#list-email-preferences" role="tab" aria-controls="email-preferences">@lang('profile.email_preferences')</a>
           <a class="list-group-item list-group-item-action" id="list-calendar-links-list" data-toggle="list" href="#list-calendar-links" role="tab" aria-controls="calendar-links">@lang('profile.calendars.title')</a>
-          @if (Auth::id() == $user->id)
-              <a class="list-group-item list-group-item-action" id="list-notifications-list" href="{{ route('notifications') }}" role="tab">@lang('profile.notifications')</a>
+          <a class="list-group-item list-group-item-action" id="list-notifications-list" href="{{ route('notifications') }}" role="tab">@lang('profile.notifications')</a>
+          @endif
+          @if(Auth::user()->isRepairDirectoryRegionalAdmin() || Auth::user()->isRepairDirectorySuperAdmin())
```
ngm, line 44: Could this be in a policy also?  So we can call e.g. @can('viewRepairDirectorySettings') rather than having the same check inline here and below.

ngm (changes requested):

app/Policies/UserPolicy.php
edwh: Sorry, yes, that will go.

app/Role.php
edwh: We do map - see below.

app/User.php
edwh: Double =, so 0 == NULL.  

It's not pleasant, but I don't want to mess with select2, really.

app/Policies/UserPolicy.php
edwh: Will change.

app/Http/Controllers/UserController.php
edwh: Will do.

app/Http/Controllers/UserController.php
edwh: Thanks, you're right - I had this try in as debug but it can now go.

resources/views/user/profile-edit.blade.php
edwh: Yes, it could and should be.  This is true of a zillion checks in the various blade templates.  

For this specific story, what do you want?

- Just this template
- All instances of this specific || test in other templates
- Policies for all related Auth checks in other templates

ngm: For this specific story, just for this template.

ngm (approved):

PR #218, opened 2020-11-20
Allow network administrator to make someone a host

edwh:
The front end code already displays the pencil appropriately.  This change modifies the permission check to allow a network coordinator to make someone a host.

It doesn't seem right to me that network coordinator is a user-level field.  You could surely be a coordinator of one network and not have that role on others.  So I may be missing something here - please review the logic.

### DIFF
--- a/app/Http/Controllers/GroupController.php
+++ b/app/Http/Controllers/GroupController.php
@@ -936,9 +936,9 @@ class GroupController extends Controller
         $group = Group::find($group_id);
         $loggedInUser = Auth::user();
 
-        if ((FixometerHelper::hasRole($loggedInUser, 'Host') && FixometerHelper::userIsHostOfGroup($group_id, Auth::id())) ||
+        if (($loggedInUser->hasRole('Host') && FixometerHelper::userIsHostOfGroup($group_id, $loggedInUser->id)) ||
             $loggedInUser->isCoordinatorForGroup($group) ||
-            FixometerHelper::hasRole($loggedInUser, 'Administrator')) {
+            $loggedInUser->hasRole('Administrator')) {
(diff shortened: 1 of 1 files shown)


ngm:
> It doesn't seem right to me that network coordinator is a user-level field. You could surely be a coordinator of one network and not have that role on others. So I may be missing something here - please review the logic.

It's a user-level field with regards to their overall role, for quick checks as to what they might see in the interface, but users are then assigned to networks via the user_network table.  

We can just use $user->isCoordinatorOf($group) method to see if the given user has rights over the given group.

app/Http/Controllers/GroupController.php
```diff
@@ -929,9 +929,16 @@ public function ajaxDeleteImage($group_id, $id, $path)
...
-        if ((FixometerHelper::hasRole(Auth::user(), 'Host') && FixometerHelper::userIsHostOfGroup($group_id, Auth::id())) || FixometerHelper::hasRole(Auth::user(), 'Administrator')) {
-            $group = Group::find($group_id);
+        // Has current logged in user got permission to add host?
+        // - Is a host of the group.
+        // - Is a network coordinator of a network which the group is in.
+        // - Is an Administrator
+        $group = Group::find($group_id);
+        $loggedInUser = Auth::user();
+
+        if (($loggedInUser->hasRole('Host') && FixometerHelper::userIsHostOfGroup($group_id, $loggedInUser->id)) ||
```
ngm, line 939: I've changed this to prefer methods that are on the user rather than in FixometerHelper or inline.  We would probably want to factor this out in to a policy in future..

ngm (approved):

PR #219, opened 2020-11-23
DOT-1332 microtasking landing page

ngm:


app/Http/Controllers/MicrotaskingController.php
```diff
@@ -0,0 +1,100 @@
...
+        $client = app('discourse-client', ['username' => $username]);
+
+        $topics = [];
+
+        try {
+            $endpoint = "/tag/{$tag}/l/latest.json";
+            $response = $client->request('GET', $endpoint);
+            $discourseResult = json_decode($response->getBody());
+
+            $topics = $discourseResult->topic_list->topics;
```
edwh, line 77: In my test environment, $topics was NULL, which the code doesn't handle.  That might be because I'm not set up right for Discourse testing.  So I can't tests this, but probably worth coding for it too.

resources/assets/js/components/DiscourseTopic.vue
```diff
@@ -0,0 +1,118 @@
...
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+}
+
+
+
+.badge-wrapper {
+    font-size: .8706em;
```
edwh, line 89: There's lots of this kind of hard-coded font size, px numbers, stuff in the codebase, and I've added some too.  Horrid, isn't it?

resources/assets/js/components/MicrotaskingCallToAction.vue
```diff
@@ -0,0 +1,76 @@
...
+</script>
+
+<style scoped lang="scss">
+@import 'resources/global/css/_variables';
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins/_breakpoints';
+
+#cta {
+    background-color: #4aaebc;
```
edwh, line 51: I think you can use $brand-light for this colour.

resources/lang/en/microtasking.php
```diff
@@ -0,0 +1,45 @@
...
+        'number_of_quests' => 'Number of quests',
+        'all_volunteer_contributions' => 'Upstream volunteer contributions',
+        'all_quests' => '3 quests',
+        'my_contributions' => 'My contributions',
+        'my_quests' => ':value quests',
+        'join' => 'Join',
+    ],
+    'cta' => [
+        'title' => 'Help Now!  MobiFix',
+        'description' => '                    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
```
edwh, line 35: Intentional placeholder text?

resources/assets/js/components/DiscourseTopic.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+    <b-tr class="topic">
+        <b-td class="main-link"><a :href="url">{{ topic.title }}</a>
+            <div>
+                <a class="badge-wrapper" :href="categoryUrl">
+                    <span class="badge-category-bg"
+                          v-bind:style="{ backgroundColor: '#'+topic.category.color}"
+></span>
```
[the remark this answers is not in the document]
ngm: Yes, this comes back from Discourse and we want to use that (they seem to seem style as well in their front-end, so mimiced that.

resources/assets/js/components/DiscourseTopic.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+    <b-tr class="topic">
```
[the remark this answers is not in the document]
ngm: Ah good tip!  Here I nabbed the markup that they're using in Discourse so I could also nab the CSS from there.

resources/views/microtasking/dashboard.blade.php
```diff
@@ -0,0 +1,36 @@
...
+        <div class="vue-placeholder vue-placeholder-large">
+            <div class="vue-placeholder-content">@lang('partials.loading')...</div>
+        </div>
+
+        <div class="vue">
+            <MicrotaskingPage
+              :total-contributions="{{ $totalContributions }}"
+              :current-user-quests="{{ $currentUserQuests }}"
+              :current-user-contributions="{{ $currentUserContributions }}"
+              :topics="{{ json_encode($topics) }}"
```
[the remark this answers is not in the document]
ngm: I think here topics is the only complex bit of data - the others are just numbers.  I guess maybe no encoding for primitive types, encoding for complex types?

app/Http/Controllers/MicrotaskingController.php
ngm: Hmm, did you get an error?  It should catch the exception, log it, and return an empty array.  (Which yes I think I've forgotten to handle on the client side with a message of 'no topics found').

resources/assets/js/components/DiscourseTopic.vue
ngm: I've taken this from the Discourse CSS again, although it is such a remarkably specific number I guess they're calculating it somewhere.  What would be better here?

resources/assets/js/components/MicrotaskingCallToAction.vue
ngm: Ah yes thanks - will use the variable.

resources/lang/en/microtasking.php
ngm: It is for now, waiting on the content from James.

resources/assets/js/components/DiscourseTopic.vue
```diff
@@ -0,0 +1,118 @@
...
+      required: true
+    },
+    discourseBaseUrl: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    ago() {
+      return new moment(this.topic.created_at).fromNow()
```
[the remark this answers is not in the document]
ngm: Don't think it matters here too much - I don't think the topics are ever so recent that they need updating in terms of seconds or minutes.  But if it's simple to add in, why not (or at least to see how, for other places where the data is more regularly refreshed)

app/Http/Controllers/MicrotaskingController.php
```diff
@@ -0,0 +1,100 @@
...
+            'currentUserQuests' => $currentUserQuests,
+            'currentUserContributions' => $currentUserContributions,
+            'topics' => $this->getDiscussionTopics($tag, 5),
+            'seeAllTopicsLink' => env('DISCOURSE_URL') . "/tag/{$tag}/l/latest"
+        ]);
+    }
+
+    private function getUserContributions($userId)
+    {
+        $faultCatContributions = DB::select('select count(*) as total from devices_faults_opinions where user_id = :userId', ['userId' => $userId])[0]->total;
```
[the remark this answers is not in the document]
ngm: Yes I'll look at this again - I don't think we have much in the way of an existing model here, so went raw, but perhaps should introduce one.

resources/assets/js/components/DiscourseTopic.vue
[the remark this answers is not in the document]
edwh: What you do is start a timer to write the time to the store regularly:

```
    updateTime() {
      this.$store.dispatch('misc/setTime')
      this.timeTimer = setTimeout(this.updateTime, 10000)
    }
```

(You'll need to clear the timer too in beforeDestroy).

In the store you have the action and getter:

```
import Vue from 'vue'

export const state = () => ({
  cdn: process.env.CDN,
  time: null
})

export const mutations = {
  set(state, params) {
    Vue.set(state, params.key, params.value)
  },

  setTime(state) {
    state.time = new Date()
  }
}

export const getters = {
  get: state => key => state[key],
  time: state => state.time
}

export const actions = {
  set({ commit }, params) {
    commit('set', params)
  },

  setTime({ commit }) {
    commit('setTime')
  }
}

```

Then in a computed property you add a fake dependency on the time

```
    timeago() {
      let ret = null

      // Make depend on misc/time so that reactivity updates.
      if (this.$store.getters['misc/time'] && this.chatmessage) {
        ret = this.$dayjs(this.chatmessage.date).fromNow()

        // dayjs pluralises wrongly in some cases - we've seen 1 hours ago.
        ret = ret.replace(this.dePlural, '1 $1')
      }

      return ret
    }

```

So when the timer updates the store, reactivity forces a re-evaluation of the computed property.

The avoids the horror of having a zillion timers running, which can kill the browser.

resources/views/microtasking/dashboard.blade.php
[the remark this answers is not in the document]
edwh: Yes, I think so, but double-check - getUserContributions returns an object so I'm a little surprised it works.

resources/assets/js/components/DiscourseTopic.vue
edwh: Probably the ideal is to introduce a set of variables, and then use those consistently throughout the code.  But I think it's a bit hopeless at the moment, it really needs a global overhaul as per the colours.

PR #221, opened 2020-11-23
REP-60 restarters menu

edwh:
This adds the Repair Directory menu item into the user menu dropdown when the logged-in user has permission to see it.

This is built on REP-58, so that needs to get released first.


ngm (approved):
Just looking at resources/views/layouts/navbar.blade.php, as everything else is from REP-58 - this looks good to me.

ngm (approved):

PR #224, opened 2020-11-27
DOT-1286 global impact

edwh:
Redesign top section


app/Device.php
```diff
@@ -625,4 +625,32 @@ public function getImages()
 
         return $File->findImages(env('TBL_DEVICES'), $this->iddevices);
     }
+
+    public function fixedPoweredCount() {
+        // We want fixed devices with an powered category.
+        $count = Device::where('repair_status', env('DEVICE_FIXED'))->withCount(['deviceCategory' => function($query) {
```
ngm, line 631: This is nice, didn't know about withCount.

ngm (approved):

PR #227, opened 2020-11-27
DOT-1357 multiple environments

edwh:
This shows the current branch prominently.  It's controlled by APP_SHOW_BRANCH, which we would only set in development environments or staging systems.

Grabbing the branch from git means we don't have to remember to edit the .env file when we switch between branches, which we would undoubtedly forget to do.

image


resources/views/layouts/navbar.blade.php
```diff
@@ -12,6 +12,22 @@
     </div>
 </a>
 
+@if(env('APP_SHOW_BRANCH'))
```
ngm, line 15: I think that it's perhaps marginally better to use the `config` helper where we can, rather than `env` helper.  In order to slightly decouple configuration from the .env file.  We'd still use the .env file for now, but makes it easier to move should we ever wish for settings to come from elsewhere, e.g. the database.

That said, certainly not a big deal here, and would be good to get your opinion on this - let's discuss at the retro.

ngm (approved):

PR #229, opened 2020-12-02
DOT-1287 repair records

edwh:
This reworks the repair records table into the new design.  Filter code is partially present but is excluded till 1288.

The layout uses Bootstrap's table, making use of the items provider to fetch the data from the API, and the built-in pagination function.

Clicking on a row brings up a modal.  This may be a temporary measure - the design has the details expanding inline, but that would require some effort to make work, since b-table is aimed at showing rows of table data, not other stuff.  The modal is full-screen to get the three column layout on desktop; if we wanted a smaller modal that might require more work to make EventDevice a bit better in terms of responsiveness at different sizes.

DeviceController::index is simplified.  DeviceController::search is no longer needed - is replaced by ApiController::getDevices, which will take optional parameters for the filtering/searching.


ngm (approved):
Looks good, just needs merging against develop.

PR #230, opened 2020-12-09
DOT-1158 Fixometer showing undefined weight for most recent event after search

edwh:
Missing bit of code in DeviceController::search().

There's duplication between search() and index(), but search() will be removed once DOT-1287 is released.


ngm (approved):

ngm:
Tested - works :+1:

PR #231, opened 2020-12-09
DOT-1367 User's calendar link returns an error

edwh:
This was a bug introduced by me in DOT-1315.  I've fixed the case causing an error, and added a similar protection in the other place it might apply (it doesn't at the moment, but is safer and clearer code).


ngm (approved):
Great - probably a good policy for us to adopt - always use fully qualified column names? (even when it might seem redundant, you never know how a query might evolve over time)

edwh:
If you're using multiple tables, yes.  It means you don't have to 
internalise which columns live in which table.

PR #234, opened 2020-12-21
Tech Debt 

edwh:
Various things in this one.

- Early on when we were incrementally migrating blade templates to Vue components, I had a pattern of putting props into a mixin, and then using $props to pass data into the component.  This meant that everything I might need was available inside components.  In retrospect that was a mistake - it hides what the interface to a component actually is.  This PR removes its use (for the group and event) mixins, by moving the relevant props into the components which use them, and passing them explicitly instead of using $props.
- We also had a TODO to think about whether some stats should be in the store.  On reflection, they should, so I've moved them there.  
- We were pulling the CSRF token from the DOM using jQuery in several places.  That is wrong, so I've moved it into the Store.
- Various other TODOs fixed/removed after consideration.

I've tested this, but it's the kind of thing that might have less obvious bugs, so we shouldn't rush it into production.


edwh:
More to come in this PR, hold off for now.

edwh:
Ok, worth having a review now.

resources/assets/js/mixins/group.js
```diff
@@ -1,14 +1,6 @@
-// This mixin includes lots of function relating to groups.
-// TODO In due course the group will move into the store and we'll just pass the id.  All the other props will then
-// become computed data in here.
+// This mixin includes unction relating to groups.
```
ngm, line 1: (Unimportant!) typo

ngm (approved):
All look like sensible changes to me.  Let's release them on Monday.  (After the data drive officially finishes on Saturday!)

PR #235, opened 2021-01-04
Prevent error if create event with no location

edwh:
I can't reproduce the original problem, but the code doesn't look like it handles missing locations well.

I've added some validation on the server to reject the create gracefully if no location is supplied and the event is not online.

The user experience is poor for errors on this form - we have basic HTML validation using "required", which means we see the default browser error message when we try to submit something which is missing.  But we don't have any more complex validation, which makes it hard to validation combinations of fields like this, except on the server.  Server-side validation means we lose the form data we submitted.  There are also other error cases (e.g. create an event with an end time after the start time).  

We should rewrite the event creation page into Vue and add better validation, and perhaps then consider whether we really need pre-moderation of events rather than post-moderation.  Let me know if you want me to create a story for that.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -182,11 +182,18 @@ class PartyController extends Controller
                 'event_date' => 'required',
                 'start' => 'required',
                 'end' => 'required',
+                'location' => [
+                    function ($attribute, $value, $fail) use ($request){
+                        if (!$request->filled('online') && !$value) {
+                            $fail(__('events.validate_location'));
+                        }
+                    }
+                ]
             ]);
 
             $error = array();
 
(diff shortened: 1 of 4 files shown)


ngm:
> Server-side validation means we lose the form data we submitted.

Technically speaking we can use Laravel's 'old input' helpers to not lose the form data - https://laravel.com/docs/5.6/requests#old-input - although it appears we aren't here.

> We should rewrite the event creation page into Vue and add better validation, and perhaps then consider whether we really need pre-moderation of events rather than post-moderation. Let me know if you want me to create a story for that.

Yes - they will be low priority for now though. And they should be two separate stories I would think.

ngm (approved):
LGTM

PR #245, opened 2021-02-03
DOT-1363 landing page

edwh:
MVP of landing page in new design.


app/Http/Controllers/DashboardController.php
```diff
@@ -151,27 +153,25 @@ public function index()
...
             'upcoming_events' => $upcoming_events,
-            'outdated_groups' => $outdated_groups,
-            'inactive_groups' => $inactive_groups,
-            'news_feed' => $news_feed,
-            'all_groups' => $all_groups,
-            'onboarding' => $onboarding,
-            'impact_stats' => $impact_stats,
-            'wiki_pages' => $wiki_pages,
-            'hot_topics' => $this->getDiscourseHotTopics(),
+            'topics' => $this->getDiscourseHotTopics()['talk_hot_topics'],
```
ngm, line 172: I believe the intention was for this to change to latest topics, as opposed to hot topics - though perhaps this comes in part 2 of these two part stories.

app/Party.php
```diff
@@ -39,7 +39,7 @@ class Party extends Model implements Auditable
         'shareable_code',
         'online',
     ];
-    protected $hidden = ['created_at', 'updated_at', 'deleted_at', 'frequency', 'group', 'group', 'idevents', 'user_id', 'wordpress_post_id'];
+    protected $hidden = ['created_at', 'updated_at', 'deleted_at', 'frequency', 'group', 'group', 'user_id', 'wordpress_post_id'];
```
ngm, line 42: What's the reason for removing this?  Is it because we need idevents in an array somewhere?

Just wondering if it might have an unintended consequence somewhere else.  I guess probably not, as it is adding rather than removing a field from what is included in arrays/json.

resources/assets/js/components/DashboardRightSidebar.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div>
+    <div class="d-none d-md-block">
+      <b-img-lazy fluid src="/images/dashboard_3.jpg" class="border border-dark border-bottom-0" />
+      <div class="pt-4 pl-4 pr-4 pb-2 greyish border border-dark border-bottom-0">
+        <p class="font-weight-bold">
+          <!--        eslint-disable-next-line-->
+          {{ translatedSidebarIntro1 }}
+          <span class="d-inline d-md-none">
+            {{ translatedSidebarKit1 }} <a href="https://talk.restarters.net/t/how-to-run-a-repair-event-the-restart-party-kit/324">{{ translatedSidebarKit2 }}</a>
```
ngm, line 10: We should review whether we need to use the special links that force login to Discourse via SSO, if not already logged in - added to retro list.

resources/assets/js/components/GroupEventSummary.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <b-tr v-if="event && stats" :class="{
```
ngm, line 2: Not sure what these changes are - are the unrelated to the landing page - or is this component now also used on the landing page?

resources/lang/de/partials.php
```diff
@@ -35,7 +35,7 @@
   'area_text_1' => 'Through this community, potential volunteers self-register and share their location. The platform is designed to connect organisers and fixers.',
   'area_text_2' => 'Through this community, potential volunteers self-register and share their location. Here\'s a list of potential volunteers near you',
   'upcoming_events' => 'Upcoming events',
-  'no_upcoming_events' => 'No Upcoming Events',
+  'no_upcoming_events' => 'No upcoming Events',
```
ngm, line 38: If it's changing to lower case, events should also be lower case, no?

resources/assets/js/components/MicrotaskingVolunteering.vue
```diff
@@ -118,15 +116,6 @@ export default {
 @import '~bootstrap/scss/variables';
 @import '~bootstrap/scss/mixins/_breakpoints';
 
-#volunteering {
```
ngm, line 121: Has this and the breakpoint difference moved somewhere else?

resources/views/dashboard/index.blade.php
```diff
@@ -79,22 +34,55 @@
...
-          @include('dashboard.restarter')
-        @endif
-        @if (FixometerHelper::hasRole($user, 'NetworkCoordinator'))
-            @include('dashboard.coordinator')
-        @endif
-        <div class="col-12">
-            @include('dashboard.blocks.impact')
-        </div>
+      <?php
+        function expandGroups($groups) {
```
ngm, line 38: I have a memory that we said we'd try to move this type of expansion logic in to the controller, and pass the expanded result in to the view, going forwards - should we?  

That said, in this case, where is expandGroups actually called?

ngm (changes requested):
Looks good, just a few questions peppered in there.

app/Party.php
edwh: Yes, exactly.

app/Http/Controllers/DashboardController.php
edwh: I did wonder about that.  

The Workbench page shows the latest ones, via MicrotaskingController::getDiscussionTopics.  Where would be a good place to move that method to?

resources/assets/js/components/GroupEventSummary.vue
edwh: Fixing some errors - we'd get JS errors if stats is null.

resources/lang/de/partials.php
edwh: That looks like a search and replace mistake - there's no reason to make this change to the German version if we're not making it to the English version.  Thanks for spotting.

resources/assets/js/components/MicrotaskingVolunteering.vue
edwh: This was a bad change.  There is a duplicate id of "volunteering" which was causing a warning, because MicrotaskingPage also defines it.  But I can't see a good reason to remove the style at all.

Given that CSS can be scoped within a component there is very little reason to use id at all now, and some reasons not to.

resources/views/dashboard/index.blade.php
edwh: I think I left this code in by mistake having moved the logic into the controller.  Removed it now.

resources/assets/js/components/DashboardRightSidebar.vue
ngm: Following discussion, we said yes we would keep them on pages that are behind login.

app/Http/Controllers/DashboardController.php
ngm: If it's logic that's going to be shared between two controllers, I'd probably create a service class, Services/DiscourseService.php, that's injected in to each of the controllers.

That said, on Workbench it is getting the latest topics for a specific tag, not the latest topics for all of Discourse which I think is the intention here, so it might need a bit of rejigging to accommodate that (or a new method).

PR #246, opened 2021-02-08
DOT-1383 Fix export

edwh:
I broke this when removing code I thought was dead.  The old code used DeviceController::search, which no longer exists.  

I'm not sure that's necessary or the best way to do it.  Can you think about whether there is any reason not to return all the devices?

### DIFF
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -55,7 +55,7 @@ class ExportController extends Controller
         // To not display column if the referring URL is therestartproject.org
         $host = parse_url(\Request::server('HTTP_REFERER'), PHP_URL_HOST);
 
-        $all_devices = app('App\Http\Controllers\DeviceController')->search($request, true);
+        $all_devices = Device::all();
 
         // Create CSV
         $filename = 'devices.csv';
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -55,7 +55,13 @@ class ExportController extends Controller
         // To not display column if the referring URL is therestartproject.org
         $host = parse_url(\Request::server('HTTP_REFERER'), PHP_URL_HOST);
 
-        $all_devices = Device::all();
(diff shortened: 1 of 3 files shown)


ngm:
> I broke this when removing code I thought was dead. The old code used DeviceController::search, which no longer exists.
> 
> I'm not sure that's necessary or the best way to do it. Can you think about whether there is any reason not to return all the devices?

The reason would be so that the export matches the way in which you have filtered the list. What's your argument for not doing that?  (One argument could be that you can filter it more easily in a spreadsheet following your export.)

In the interest of getting the export working again, to begin with let's just re-enable it as a full export if this is quicker.  Given where the button currently is, it doesn't indicate that the export is filtered.

Then, in a future sprint, we can determine whether there is any utility in a filtered export.

edwh:
Then let's do a full export.  Are you OK with the code?

ngm (changes requested):
The code for the export looks good, can you confirm check though that the Export Data button is only present for Admins?  (as per in the Jira ticket)

edwh:
The Export button is there for everyone.  I've not changed that.

We could make it only visible for Admins, but I'm not sure it's part of 
this story, and I'm not sure there's much point given that anyone who 
cares can grab the data in other ways?

ngm:
> We could make it only visible for Admins, but I'm not sure it's part of this story, and I'm not sure there's much point given that anyone who cares can grab the data in other ways?

From the midsts of time I recall that via the download from https://therestartproject.org/download-dataset, you get the data without the model field.  Internally you get the full dataset, including the model field.  The reason (perhaps no longer valid and to be discussed), was the desire to keep the model field private as a potential business asset.  

So I think we lost the fact that the Export Data button was for admins only when we recently migrated this page to the new design.  Until we have a solidly groomed story on the export from the Fixometer page for non-admins, we should keep it for admins only.

edwh:
Ok - change pushed.  Have a look?

ngm:
@edwh Looks good :+1:

PR #251, opened 2021-02-17
DOT-1391 Insert modal for Add Data

edwh:
As discussed, the modal already exists and (pretty much) works if we insert it.

### DIFF
--- a/app/Http/Controllers/DeviceController.php
+++ b/app/Http/Controllers/DeviceController.php
@@ -49,8 +49,17 @@ class DeviceController extends Controller
                             ->homepage_data();
         $global_impact_data = $global_impact_data->getData();
 
+        $user_groups = [];
+
+        if ($user) {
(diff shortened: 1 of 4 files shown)


ngm (commented):
Noting that the link from 'Go to Event' goes to an anchor named #devices-section, but the anchor no longer exists of the events page.  

The idea being to take you to the part of the page where you'll be adding the data. 

Out of interest, can we reinstate the anchor?  Does that work in a simple way with Vue?

edwh:
It doesn't work in a very simple way when using Vue in the fragmented way in which we are here.

If we had a full app, we'd have SSR and then scroll to the relevant element in mounted(), and we could make that global behaviour via a mixin.

In our case it's tricky, because we don't know exactly when everything has finished rendering and the page is stable.  So I've added a dumb scroll after 2 seconds which will work a reasonable proportion of the time.

ngm:
@edwh Tested on five random events, and it worked for me each time.  Good to go then!

PR #252, opened 2021-02-17
DOT-1390 Remove cookies for stats

edwh:
- Create a new 'nocookie' layout, which doesn't include the Google stuff.
- Remove the cookie notice which is then not needed
- Use this from the stats template.
- Stop the global JS barfing.

Tested in Brave to ensure it looks clean.

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -999,17 +999,20 @@ function initAutocomplete() {
 
   jQuery(document).ready(function () {
     groupsMap();
-    gdprCookieNotice({
-      locale: 'en',
-      timeout: 500, //Time until the cookie bar appears
-      expiration: 30, //This is the default value, in days
-      domain: restarters.cookie_domain, //If you run the same cookie notice on all subdomains, define the main domain starting with a .
-      implicit: false, //Accept cookies on page scroll automatically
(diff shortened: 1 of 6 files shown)


resources/assets/js/app.js
```diff
@@ -999,17 +999,20 @@ function initAutocomplete() {
...
-      expiration: 30, //This is the default value, in days
-      domain: restarters.cookie_domain, //If you run the same cookie notice on all subdomains, define the main domain starting with a .
-      implicit: false, //Accept cookies on page scroll automatically
-      statement: '/about/cookie-policy', //Link to your cookie statement page
-      performace: ['DYNSRV'], //Cookies in the performance category.
-      analytics: ['_ga','_gat', '_gid'], //Cookies in the analytics category.
-      marketing: [] //Cookies in the marketing category.
-    });
+
+    if (window.gdprCookieNotice) {
```
ngm, line 1003: When does this evaluate to true?  Is it a check for the presence of the function?

ngm (approved):

resources/assets/js/app.js
edwh: Yes.  There are umpteen ways of doing this; possibly https://stackoverflow.com/a/60511084/775258 would be better.  But I think this will do in the global JS stuff.

PR #254, opened 2021-02-24
DOT-1134 landing page part 2

edwh:
Update to the new design.


app/Services/DiscourseService.php
```diff
@@ -0,0 +1,47 @@
...
+
+            foreach ($topics as $topic) {
+                foreach ($categories as $category) {
+                    if ($topic->category_id == $category->id) {
+                        $topic->category = $category;
+                    }
+                }
+            }
+        } catch (\Exception $ex) {
+            Log::error("Error retrieving microtasking discussion topics for username '{$username}': " . $ex->getMessage());
```
ngm, line 42: No longer only microtasking related topics.

ngm (approved):
Noticed that a number of the translations are still using the translatedX computed properties, not the new __() method.  No need to go through and change them all here.. just wondering why? (probably because it was introduced in a commit after this pull request was started, I'm guessing)

edwh:
> Noticed that a number of the translations are still using the translatedX computed properties, not the new __() method. No need to go through and change them all here.. just wondering why? (probably because it was introduced in a commit after this pull request was started, I'm guessing)

Just to be consistent until we do the story to go through and move them over.

app/Services/DiscourseService.php
edwh: Ah yes.

PR #255, opened 2021-03-08
DOT-1289 fixometer edit

edwh:
This changes the fixometer table to allow you to edit/delete.  It uses Bootstrap-Vue's row details feature, which I've only just discovered, to insert the row in the right place.  The popup modal for viewing details and the individual device view/edit page are retired.

There's a bit of other tidyup/error fixing in here.

The layout isn't great - because we have a narrower width here the three columns for the device wrap.  But this is a rare case for admins only, so I don't think we need to worry about that.


ngm (commented):
Code looks good in general, have left some functional review notes in Jira.

PR #256, opened 2021-03-12
Introduce a better translation mechanism

edwh:
We currently do translations in Vue components by having computed properties which call this.$lang.get().  This works but leads to a lot of dull code.

This PR introduces a `__()` method which is in a global mixin and therefore available everywhere.  This is named to match the Laravel translation call, which means it's one fewer thing to remember.  We can then pass in the key directly in a template, e.g. 

```
            <p>
              {{ __('dashboard.catch_up') }}
            </p>
```

I've used it in one example, but if you like this approach then I'll go through and apply it throughout - bit dull to do but removes a bunch of code that doesn't deserve to live.

In Vue 2, filters might be a natural way to do this, and I like them, but they're being retired in Vue 3.  

This has the additional minor advantage of making it easier to refactor by moving template chunks between components - at the moment you also have to move the relevant computed properties, and we can easily end up with computed properties which are no longer actually used.

### DIFF
--- a/resources/assets/js/app.js
+++ b/resources/assets/js/app.js
@@ -1517,11 +1517,19 @@ function initAutocomplete() {
 jQuery(document).ready(function () {
   // Vue.
   //
-  // Create a mixin so that $lang is available in all components.
+  // Create a mixin for translation.
   Vue.mixin({
     computed: {
       $lang() {
+        // We want this to be available in all components.
         return Lang
+      }
+    },
+    methods: {
+      __(key) {
+        // This means we can use __('key') in Vue templates in the same way as we are used to in Laravel
+        // templates.
+        return this.$lang.get(key)
       }
     }
   })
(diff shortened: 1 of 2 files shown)


ngm (approved):
Great stuff, I like it!  Will remove a lot of boilerplate from the components.

edwh:
Ok.  Shall I merge this and then create a story for the application 
throughout?

I think there might be some other stuff left, e.g. "choice" - but this 
is most of it.

From: "Neil M" ***@***.***>
To: "TheRestartProject/restarters.net" 
***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Author" 
***@***.***>
Sent: 15/03/2021 10:11:15
Subject: Re: [TheRestartProject/restarters.net] Introduce a better 
translation mechanism (#256)

>
***@***.*** approved this pull request.

ngm:
> Ok. Shall I merge this and then create a story for the application throughout? I think there might be some other stuff left, e.g. "choice" - but this is most of it.

Yes please, merge this in with just the existing change, and then create a story for the remaining changes.

PR #257, opened 2021-03-15
DOT-1396

edwh:
This changes the server to query Discourse anonymously if you're not logged in, and then to always display results in on the client.

Also fixes a missing computed property (not tracked down where that snuck in), and some formatting.


ngm (approved):
All looks good.   Looks like a bit of overlap with 1134 to be merged together.

Noting (for myself) that we now have a separate api client specifically for use when we want to access Discourse API on behalf of anonymous user.

PR #260, opened 2021-03-19
DOT-1405 Change navbar for microtasks

edwh:
There's already a flag for this - we just need to use it in the appropriate index.php blades.


ngm (changes requested):
Ah good, that's simple then.

Sorry, forget to mention - each of the tasks has a /status page that needs this change too.

e.g. /mobifixora/status

edwh:
Applied to those too.

ngm (approved):

PR #268, opened 2021-04-07
DOT-1415 Group name lost from Fixometer URL

edwh:
The smallest of fixes - one character typo.

### DIFF
--- a/resources/assets/js/components/FixometerPage.vue
+++ b/resources/assets/js/components/FixometerPage.vue
@@ -389,7 +389,7 @@ export default {
       }
 
       if (this.group) {
-        ret += 'group=' + encodeURIComponent(this.groupd) + '&'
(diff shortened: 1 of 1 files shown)


ngm (approved):
Complete tangent, but I liked the quasar feature where slaps in your face a huge error during dev if there's anything wrong with your JS (even linting stuff).  As someone not well-versed in JS or Vue I found it very handy.  I imagine nuxt does/can do similar, and that we probably can't do similar in Restarters, because of our hybrid setup.

edwh:
Yes, if you use nuxt the lint stuff throws an error by default, and you can use auto-fix to correct stuff too.  It helps a lot.  Everything is possible, so I presume there would be a way to get it to work with the hybrid stuff, but I'd rather invest the effort in moving towards a real client/server split.

PR #269, opened 2021-04-07
DOT-1413 online event badge

edwh:
Introduce an EventTitle component which shows a badge when the event is online.

The eslint disable is to prevent IDEs (like mine) auto-formatting the HTML in a way that introduces a space, which then becomes part of the link and gets underlined.

### DIFF
--- a/resources/assets/js/components/EventTitle.vue
+++ b/resources/assets/js/components/EventTitle.vue
@@ -1,0 +1,30 @@
+<template>
+  <component :is="component">
+    {{ event.venue ? event.venue : event.location }}
+    <b-badge v-if="event.online" variant="primary" pill>
+      {{ translatedOnline }}
+    </b-badge>
+  </component>
+</template>
+<script>
+export default {
+  props: {
+    idevents: {
+      type: Number,
+      required: true
+    },
+    component: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    event() {
+      return this.$store.getters['events/get'](this.idevents)
+    },
+    translatedOnline() {
+      return this.$lang.get('events.online')
+    }
+  }
+}
+</script>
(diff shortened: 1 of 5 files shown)


resources/assets/js/components/EventTitle.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <component :is="component">
```
ngm, line 2: Ah this interesting, it's a way to make a component template be configured to use different html elements?

resources/assets/js/components/EventTitle.vue
```diff
@@ -0,0 +1,33 @@
...
+      required: true
+    },
+    component: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    event() {
+      return this.$store.getters['events/get'](this.idevents)
```
ngm, line 21: General Vue question - is it preferable to pass in an id to a component, then get the associated entry from the store, rather than passing in the entity itself?  (or depends on the context?)

ngm (approved):

resources/assets/js/components/EventTitle.vue
edwh: Yes, exactly - not just native HTML elements, though, it works with different Vue components.

resources/assets/js/components/EventTitle.vue
edwh: Yes.  I no longer remember exactly why, but I started off passing objects around and shifted over time to passing ids around.  

I think there may be some issues with reactivity, or possibly just performance.  Either it doesn't kick in, or what happens is that a small change to the object results in the whole child component being re-rendered because the prop has changed, whereas with the id approach it would just trigger re-render of the relevant bits of the DOM inside the child component.

Discuss at retro if not clear.

PR #277, opened 2021-04-14
DOT-1411 improve translation

edwh:
Removes the simple translation computed properties in favour of the `__()` method.


ngm (approved):

ngm:
Didn't check every item, but from a random spot check of a few, can see that instances of translatedX methods where lang.get was being used have been replaced.  Those where lang.choice were being used have been kept as is.  So looks good and we'll test while in develop.

PR #281, opened 2021-04-16
DOT-740 item type

edwh:
1. Add a flag on the category and a migration to set it.
2. Return item types from the server and make available everywhere (including in places we don't use them yet but may do later, for consistency).
3. Expose the item_type field with a typeahead.
4. Clear it when the category changes.


resources/assets/js/components/EventDevice.vue
```diff
@@ -173,6 +178,30 @@ export default {
...
+        return true
+      }
+
+      let ret = false
+
+      this.clusters.forEach((cluster) => {
+        let categories = []
+
+        cluster.categories.forEach((c) => {
+          console.log("Check category", this.currentCategory, c.idcategories, c.aggregate, c)
```
ngm, line 196: Remove now?

resources/assets/js/components/EventDevice.vue
```diff
@@ -336,6 +369,10 @@ export default {
       })
 
       window.location = '/fixometer'
+    },
+    categoryChange() {
+      // Any item type we might have is no longer value.
```
ngm, line 374: value -> valid

ngm (approved):
Great stuff.  Really like the way this works :+1: 

As mentioned in Jira, currently with Ugo for some further functional testing.

PR #282, opened 2021-04-19
DOT-1418 area info

edwh:
Return the area code on group/event API calls.

### DIFF
--- a/app/Http/Controllers/API/EventController.php
+++ b/app/Http/Controllers/API/EventController.php
@@ -87,7 +87,7 @@ class EventController extends Controller
                  'value' => $party->location,
                  'latitude' => $party->latitude,
                  'longitude' => $party->longitude,
-                 'area' => $group->area,
+                 'area' => $group['area'],
              ],
              'description' => $party->free_text,
              'user' => $party_user = collect(),
--- a/app/Http/Controllers/API/EventController.php
+++ b/app/Http/Controllers/API/EventController.php
@@ -55,6 +55,7 @@ class EventController extends Controller
            $groups_array->push([
                'id' => $group->idgroups,
                'name' => $group->name,
+               'area' => $group->area,
                'description' => $group->free_text,
                'image_url' => $group->groupImagePath(),
                'volunteers' => $group->volunteers,
--- a/app/Http/Controllers/API/EventController.php
+++ b/app/Http/Controllers/API/EventController.php
@@ -55,7 +55,6 @@ class EventController extends Controller
            $groups_array->push([
                'id' => $group->idgroups,
                'name' => $group->name,
-               'area' => $group->area,
                'description' => $group->free_text,
                'image_url' => $group->groupImagePath(),
                'volunteers' => $group->volunteers,
@@ -88,6 +87,7 @@ class EventController extends Controller
                  'value' => $party->location,
                  'latitude' => $party->latitude,
                  'longitude' => $party->longitude,
+                 'area' => $group->area,
              ],
              'description' => $party->free_text,
              'user' => $party_user = collect(),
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/API/EventController.php
```diff
@@ -78,6 +79,7 @@ public function getEventsByUsersNetworks(Request $request, $date_from = null, $d
            $collection->push([
              'id' => $party->idevents,
              'group' => [$group],
+             'area' => $group['area'],
```
ngm, line 82: I think this would be better placed in the `location` array a few lines down.

app/Http/Controllers/API/GroupController.php
```diff
@@ -85,6 +85,7 @@ public static function getGroupsByUsersNetworks(Request $request)
             $collection->push([
                 'id' => $group->idgroups,
                 'name' => $group->name,
+                'area' => $group->area,
```
ngm, line 88: Same as above, I think this should go in the `location` array.

app/Http/Controllers/GroupController.php
```diff
@@ -1310,6 +1310,7 @@ public function getGroupsByKey(Request $request, $api_token)
                 $collection->push([
                     'id' => $group->idgroups,
                     'name' => $group->name,
+                    'area' => $group->area,
```
ngm, line 1313: Same as above - should go in the location array.

app/Http/Controllers/GroupController.php
```diff
@@ -1419,6 +1420,7 @@ public function getGroupByKeyAndId(Request $request, $api_token, Group $group, $
         $collection = collect([
             'id' => $group->idgroups,
             'name' => $group->name,
+            'area' => $group->area,
```
ngm, line 1423: Same as above - should go in the location array.

app/Http/Controllers/PartyController.php
```diff
@@ -1629,6 +1630,7 @@ public function getEventByKeyAndId(Request $request, $api_token, Party $party)
             'start_time' => $party->start,
             'end_time' => $party->end,
             'name' => $party->venue,
+            'area' => $party->theGroup->area,
```
ngm, line 1633: Same as above - should go in the location array.

app/Http/Controllers/PartyController.php
```diff
@@ -1616,6 +1616,7 @@ public function getEventByKeyAndId(Request $request, $api_token, Party $party)
             'group' => [
                 'id' => $party->theGroup->idgroups,
                 'name' => $party->theGroup->name,
+                'area' => $party->theGroup->area,
```
ngm, line 1619: Same as above - should go in the location array.

ngm (changes requested):
The area should go in the location array in the various places where the data is returned.

app/Http/Controllers/API/EventController.php
ngm: Looks like this one still needs moving to the location array (line 85)

ngm (changes requested):

app/Http/Controllers/API/EventController.php
```diff
@@ -88,6 +87,7 @@ public function getEventsByUsersNetworks(Request $request, $date_from = null, $d
                  'value' => $party->location,
                  'latitude' => $party->latitude,
                  'longitude' => $party->longitude,
+                 'area' => $group->area,
```
ngm, line 90: Coming back as null - needs to be $group['area'] here I believe.

ngm (changes requested):

app/Http/Controllers/API/EventController.php
edwh: Needs to be in both sections.  Try now?

ngm (approved):

PR #292, opened 2021-05-05
DOT-1439_Migrate audit columns to longtext

edwh:
The migration is a bit unusual because of bugs.

https://stackoverflow.com/questions/37724150/laravel-migrations-change-a-column-type-from-varchar-to-longtext


ngm (approved):
Ha, as you say the migration looks odd, but I can see that it results in the field changing to longtext :+1:

PR #293, opened 2021-05-05
DOT-843 Allow duplication of events

edwh:
1. Introduce a route to duplicate an event and return the Create page with that data
2. Link to it from the View/Edit event pages.


ngm (changes requested):
Just needs 

```
<EventTimeRangePicker starttimeinit="{{ old('start') }}" endtimeinit="{{ old('end') }}" />
```

change, as per Jira.

PR #296, opened 2021-05-14
DOT-1369 groups events devices update

edwh:
This introduces an event for devices being created/updated, listens on it, and uses that to maintain a devices_updated_at field in the party and group.

### DIFF
--- a/app/Device.php
+++ b/app/Device.php
@@ -2,6 +2,7 @@
 
 namespace App;
 
+use App\Events\DeviceCreatedOrUpdated;
 use DB;
 use Illuminate\Database\Eloquent\Model;
 use OwenIt\Auditing\Contracts\Auditable;
@@ -25,6 +26,16 @@ class Device extends Model implements Auditable
      * @var array
      */
     protected $hidden = [];
+
+    /**
+     * The event map for the model.
+     *
+     * @var array
+     */
+    protected $dispatchesEvents = [
+        'updated' => DeviceCreatedOrUpdated::class,
+        'created' => DeviceCreatedOrUpdated::class,
+    ];
 
     public static function boot()
     {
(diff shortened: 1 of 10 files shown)


ngm (commented):
Looks great, the event approach is very nice and clean.

Doesn't look like this is exposed to the via the APIs yet - presume we're waiting to do this as per the ticket, following clarficiation from RT developers as to whether they are in a position to use a new field.

ngm:
Aha, I was looking at the API endpoints:

/api/groups/network?api_token=x
/api/events/network?api_token=x

As these are what Repair Together access.  I don't seem to see it there even after adding a device.  

That said, after chatting with RT, they would prefer to use the existing updated_at field, and have it represent updates to either group/event properties OR device updates.  So we should return updated_at from these API calls as the greater of updated_at and devices_updated_at.  (This would avoid us having to run a script to set devices_updated_at for all groups/events too.)

edwh:
Ok - I've made those two API calls return the max value.  See what you think.

ngm:
@edwh Code all looks good, just one odd thing for me - when running locally, I get the following returned:

image

created_at is an empty array for some reason, and updated_at is a string.

On production at present, we get the following returned:

image

both created_at and updated_at are objects with properties of date, timezone_type, and timezone.

I'm not sure why that is - presumably that's how a Carbon datetime is serialized.  I don't know why this is happening one way on prod, but not locally - only think I can think is something has changed between PHP versions?

I think wrapping the dates e.g. new \Carbon\Carbon($group->max_updated_at_devices_updated_at) might preserve the existing behaviour.

edwh:
That is a bit odd.  Fix pushed as you suggest.

I tend to return dates in ISO 8601, as I think that's easier for a client to parse.  But we should preserve the existing behavour.

ngm:
> I tend to return dates in ISO 8601, as I think that's easier for a client to parse. But we should preserve the existing behavour.

Agreed (to both points).

ngm (approved):

PR #297, opened 2021-05-14
DOT-1446 regression

edwh:
Add some very very basic tests.

I've started to add some utility methods inside TestCase so that we don't duplicate code to e.g. create a user or a group.

I had to make a couple of minor fixes to get this to run - you'll see them in the diffs.

### DIFF
--- a/database/migrations/2021_05_14_173108_repairdir_role.php
+++ b/database/migrations/2021_05_14_173108_repairdir_role.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class RepairdirRole extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        if (!Schema::hasColumn('users', 'repairdir_role')) {
+            Schema::table('users', function (Blueprint $table) {
+                $table->smallInteger('repairdir_role');
+            });
+        }
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
+    }
+}
--- a/tests/Feature/Dashboard/BasicTest.php
+++ b/tests/Feature/Dashboard/BasicTest.php
@@ -8,7 +8,7 @@ use Mockery;
 use Tests\TestCase;
 
 class BasicTest extends TestCase {
-    public function testDashboardPageLoads() {
+    public function testPageLoads() {
         // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
         // Vue component exists.
         $this->loginAsTestUser();
--- a/tests/Feature/Dashboard/BasicTest.php
+++ b/tests/Feature/Dashboard/BasicTest.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Tests\Feature\Dashboard;
+
+use DB;
+use Hash;
+use Mockery;
+use Tests\TestCase;
+
+class BasicTest extends TestCase {
+    public function testDashboardPageLoads() {
+        // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
+        // Vue component exists.
+        $this->loginAsTestUser();
+        $response = $this->get('/dashboard');
+        $content = $response->getContent();
+
+        $this->assertNotFalse(strpos($content, '<DashboardPage'));
+    }
+}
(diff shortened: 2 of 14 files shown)


database/migrations/2021_05_14_173108_repairdir_role.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class RepairdirRole extends Migration
```
ngm, line 7: Hah, this raises an interesting question about migrations from the Repair Directory that change Restarters.  We should chat about it at retro.

tests/Feature/Dashboard/BasicTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+class BasicTest extends TestCase {
+    public function testPageLoads() {
+        // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
+        // Vue component exists.
+        $this->loginAsTestUser();
+        $response = $this->get('/dashboard');
+        $content = $response->getContent();
+
+        $this->assertNotFalse(strpos($content, '<DashboardPage'));
```
ngm, line 18: Think you could use assertSee here if you wanted (https://laravel.com/docs/5.6/http-tests#assert-see)

tests/Feature/Fixometer/BasicTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+class BasicTest extends TestCase {
+    public function testPageLoads() {
+        // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
+        // Vue component exists.
+        $this->loginAsTestUser();
+        $response = $this->get('/fixometer');
+        $content = $response->getContent();
+
+        $this->assertNotFalse(strpos($content, '<FixometerPage'));
```
ngm, line 18: Same as above

tests/Feature/Groups/BasicTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+class BasicTest extends TestCase {
+    public function testPageLoads() {
+        // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
+        // Vue component exists.
+        $this->loginAsTestUser();
+        $response = $this->get('/group');
+        $content = $response->getContent();
+
+        $this->assertNotFalse(strpos($content, '<GroupsPage'));
```
ngm, line 18: Same as above

tests/Feature/Microtasks/BasicTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+class BasicTest extends TestCase {
+    public function testMicrotasksPageLoads() {
+        // Test the dashboard page loads.  Most of the work is done inside Vue, so a basic test is just that the
+        // Vue component exists.
+        $this->loginAsTestUser();
+        $response = $this->get('/workbench');
+        $content = $response->getContent();
+
+        $this->assertNotFalse(strpos($content, '<MicrotaskingPage'));
```
ngm, line 18: As above

tests/Feature/Users/Registration/AccountCreationTest.php
```diff
@@ -0,0 +1,63 @@
...
+use Hash;
+use Mockery;
+use Tests\TestCase;
+
+class AccountCreationTests extends TestCase {
+    public function testRegister() {
+        $response = $this->post('/user/register/',  $this->userAttributes());
+
+        $response->assertStatus(302);
+        $response->assertRedirect('dashboard');
```
ngm, line 15: Could be worth a $this->assertDatabaseHas('users', [
        'email' => 'sally@example.com'
    ]);

to provide some basic test that the user is being created?

tests/TestCase.php
```diff
@@ -2,23 +2,103 @@
...
+        $redirectTo = $response->getTargetUrl();
+        $this->assertNotFalse(strpos($redirectTo, '/group/edit'));
+        $p = strrpos($redirectTo, '/');
+        $idgroups = substr($redirectTo, $p + 1);
+
+        return $idgroups;
+    }
+
+    public function createJane() {
+        $user = factory(User::class)->create();
```
ngm, line 94: It amounts to the same thing, but for reference you can pass values you wish to override from the default factory method with an array of attributes passed to create.

https://laravel.com/docs/5.6/database-testing#persisting-models

ngm (approved):

tests/Feature/Dashboard/BasicTest.php
edwh: Will do.

tests/Feature/Users/Registration/AccountCreationTest.php
edwh: Added.  

As a habit I tend to avoid dipping into the internals to test thing when we're trying to test the interface, as it can lead to you doing that just because the interface is a bit awkward, and then ending up with weaker interface testing.

PR #301, opened 2021-05-21
DOT-1436 event talk

edwh:
This creates a Discourse thread when an event is approved, saves the thread id, and then adds a user to that thread when they RSVP.

We need to pass a host to the Discourse API as the acting user, rather than using the global Discourse API username.


app/Party.php
```diff
@@ -62,6 +63,7 @@ public function findAll()
                     `e`.`free_text`,
                     `e`.`hours`,
                     `e`.`wordpress_post_id`,
+                    `e`.`discourse_thread`,
```
ngm, line 66: Out of interest, I wonder if some/all of these old DB::raw queries are dead code now.

ngm (approved):

app/Party.php
edwh: I wondered that too, but I was feart.

ngm: Actually quick scan says I think they might be being used.  Ho hum.  Might be a nice one for a little coverage-increasing and refactoring if I'm bored of an eve.

PR #303, opened 2021-05-26
DOT-1450 automated tests

edwh:
Add a bit of infrastructure to test Vue properties.

Make the existing CreateEventTest run.  It looks like the model for email notifications changed when network coordinators were introduced.

This increases coverage to the still risible value of 11%.

https://coveralls.io/builds/40043601


composer.json
```diff
@@ -32,6 +32,7 @@
         "soundasleep/html2text": "^1.1",
         "spatie/calendar-links": "^1.2",
         "spinen/laravel-discourse-sso": "^1.3",
+        "symfony/dom-crawler": "^5.2",
```
ngm, line 35: Should this not be in `require-dev` ?  (i.e. only needed for tests?)

edwh: Yes - pushed.

ngm (approved):

PR #306, opened 2021-05-26
DOT-1438 Switch brand field to use vue-multi-select component

edwh:
As title.  Also adds a warning if you're creating a new brand.


resources/assets/js/app.js
```diff
@@ -10,6 +10,7 @@ require('./bootstrap-sortable.js');
 require('select2');
 require('slick-carousel');
 require('summernote');
+require('summernote-cleaner');
```
ngm, line 13: From other ticket

ngm (approved):
Code all looks good :+1: 

Minor point - we don't usually check in `public/js/app.js` unless on prod, do we?

resources/assets/js/app.js
edwh: Ah yes.  Removed.

edwh:
> Minor point - we don't usually check in `public/js/app.js` unless on prod, do we?

No - must have had a search and replace.  Backed out that file change for neatness.

PR #314, opened 2021-05-28
DOT-1399 Don't count cancelled events in stats

edwh:
This seems to be straightforward:
* There's a `cancelled` property in the DB - expose that as hidden in Laravel
* There's a method for calculating the volunteer hours - adjust that.
* There's one place where we're not using that method but should be - do so.
* The group and network stats come from the event stats, so no work required.


app/Http/Controllers/AdminController.php
```diff
@@ -237,7 +237,7 @@ public function eventsCsv()
             }
 
             $participants += $party->pax;
-            $hours_volunteered += (($party->volunteers > 0 ? $party->volunteers * 3 : 12 ) + 9);
+            $hours_volunteered += $party->hoursVolunteer();
```
ngm, line 240: Typo?  Looks like it's hoursVolunteered()

app/Party.php
```diff
@@ -705,16 +705,21 @@ public function devices()
 
     public function hoursVolunteered()
     {
-        $lengthOfEventInHours = 3;
-        $extraHostHours = 9;
-        $hoursIfNoVolunteersRecorded = 12;
+        if (!$this->cancelled) {
+            $lengthOfEventInHours = 3;
```
ngm, line 709: Noting to ourselves that we must put a story in about correctly calculating the length of an event for this (knowing that we need to check how this will affect stats everywhere)

app/Party.php
```diff
@@ -705,16 +705,21 @@ public function devices()
 
     public function hoursVolunteered()
```
ngm, line 706: Don't have to add one in now, but as a side-note, this to me is the kind of thing that benefits from a pure unit test.

ngm: i.e. something that is purely business logic.  And that could also do with a little documentation/reassurance around some of the assumptions built in to it.

ngm (changes requested):
Just looks like a change needed to the call to hoursVolunteered

app/Http/Controllers/AdminController.php
edwh: Yes.

PR #319, opened 2021-06-04
DOT-1404 filter events

edwh:
Quite a large change.

First, rework the events table from b-table-simple to b-table.  
* Previously GroupEventsTableHeading and GroupEventSummary were used to display a row in the table.  We have to have a slot per column, so add components for each of these and retire those components.
* Use v-bind where we need to pass something from the parent into the column slot (e.g. the stats() method).
* Use callbacks to set the row and table classes to highlight events we are attending and the invalid values.
* Add a filters component and use it, similarly to the groups table, and enable this for the Other->All tab.
* Add a function to sort on the columns in the right way.

The code went through quite a lot of iterations to get to this point, so it's worth checking for hangover stuff that no longer seems to have any point.

I had some trouble with migrations - tweaked to make it run.


resources/assets/js/components/GroupEventSummary.vue
```diff
@@ -1,178 +0,0 @@
-<template>
```
ngm, line 1: This file just seems to be empty now - should it be deleted?

ngm: Oops, I just misread the diff.  I can see it has been deleted.

ngm (approved):
Code all looks good! :+1: 

Some functional bugs reported in Jira.

PR #321, opened 2021-06-07
DOT-1444 language sync to wiki

edwh:
As discussed, a cheap and cheerful command to sync the languages.

### DIFF
--- a/app/Providers/ScheduleServiceProvider.php
+++ b/app/Providers/ScheduleServiceProvider.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace App\Providers;
+
+use Illuminate\Support\ServiceProvider;
+use Illuminate\Console\Scheduling\Schedule;
+
+class ScheduleServiceProvider extends ServiceProvider
+{
+    public function boot()
+    {
+        $this->app->booted(function () {
+            $schedule = $this->app->make(Schedule::class);
+            $schedule->command('language:sync')->everyFiveMinutes();
+        });
+    }
+
+    public function register()
+    {
+    }
+}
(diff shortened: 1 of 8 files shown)


app/Providers/ScheduleServiceProvider.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace App\Providers;
+
+use Illuminate\Support\ServiceProvider;
+use Illuminate\Console\Scheduling\Schedule;
+
+class ScheduleServiceProvider extends ServiceProvider
```
ngm, line 8: Is this auto-registered somehow?

ngm (approved):
Looks good to me - just wondering how we ScheduleServiceProvider is registered (I don't remember how this works, so perhaps it's just automagically registered by virtue of being in the right folder)

Also - does it need a cron job to cause the schedule to run every 5 minutes, or Laravel schedules it itself?

app/Providers/ScheduleServiceProvider.php
edwh: No, it's registerd in app.php around line 173.

edwh:
> Also - does it need a cron job to cause the schedule to run every 5 minutes, or Laravel schedules it itself?

It does need a cron job.  When I added a scheduled task to the Repair Directory I added this on production:

`* * * * * cd /srv/users/serverpilot/apps/repairdirectory && php artisan schedule:run >> /tmp/laravel_schedule.out 2>&1`

So we'd do something similar, and tweak the output file name.

app/Providers/ScheduleServiceProvider.php
ngm: Ah right, yes, missed that somehow :+1:

PR #324, opened 2021-06-09
DAT-529 Use empty strings rather than nulls for empty problems.

edwh:
Add a migration to sort out the existing values, and a mutator to ensure no new null values sneak in.

### DIFF
--- a/app/Device.php
+++ b/app/Device.php
@@ -711,4 +711,9 @@ AND devices.event = events.idevents ';
         $types = DB::table('devices')->whereNotNull('item_type')->select('item_type', DB::raw('COUNT(*) as count'))->groupBy('item_type')->orderBy('count', 'desc')->get()->toArray();
         return $types;
     }
+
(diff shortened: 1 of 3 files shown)


app/Device.php
```diff
@@ -711,4 +711,9 @@ public static function getItemTypes() {
         $types = DB::table('devices')->whereNotNull('item_type')->select('item_type', DB::raw('COUNT(*) as count'))->groupBy('item_type')->orderBy('count', 'desc')->get()->toArray();
         return $types;
     }
+
+    public function setProblemAttribute($value) {
+        // Map null values to empty strings to avoid Metabase problems.
+        $this->attributes['problem'] = $value === NULL ? '' : $value;
```
ngm, line 717: Could be null coalesce operator no

`$this->attributes['problem'] = $value ?? '';

edwh: Yes; I'm too used to working on old PHP versions, clearly.  Added.

PR #327, opened 2021-06-14
DOT-1458 use group address

edwh:
Move the address into a Vue component, pass the group address, and add a link button to use it.

The original code went to some effort to set up some inputs for the address components, which I initially replicated in this branch.  But looking at the code, it only ever uses the location field, and I think the others are there purely for debug.  So I've removed them, but please consider that.


resources/assets/js/components/VenueAddress.vue
```diff
@@ -0,0 +1,79 @@
...
+      <b-btn variant="link" size="sm" v-if="groupLocation" @click="useGroup" class="pl-0" :disabled="currentValue === groupLocation">
+        {{ __('events.field_venue_use_group') }}
+      </b-btn>
+    </div>
+  </div>
+</template>
+<script>
+import Vue from 'vue'
+import VueGoogleAutocomplete from 'vue-google-autocomplete'
+import UniqueId from 'vue-unique-id';
```
ngm, line 30: What is vue-unique-id for?

ngm (approved):
Good to remove the hidden inputs - as far as I can see, they were not being used for anything.

resources/assets/js/components/VenueAddress.vue
edwh: You need an id on an `input `in order to be able to reference it using `for `with a `label`.  

But you want all the ids on the page to be unique.  This can be a problem with Vue, because it becomes quite easy to end up with duplicate ids by adding multiple copies of a component.  

vue-unique-id helps solve this by allowing you to easily generate a globally unique id.

PR #328, opened 2021-06-14
DOT_810- Show upcoming events to hosts of the event even if they are not approved

edwh:
Needs some careful review of the Eloquent join.


app/Party.php
```diff
@@ -382,10 +383,16 @@ public function attendees()
...
+      // We want to show approved events (wordpress_post_id IS NOT NULL) or where we are a host (e.g. because
+      // we created it.
       if( $by_event ) {
         return $this->join('groups', 'groups.idgroups', '=', 'events.group')
                      ->join('events_users', 'events_users.event', '=', 'events.idevents')
-                     ->whereNotNull('events.wordpress_post_id')
+                     ->where(function($query) {
+                         $query
+                         ->whereNotNull('events.wordpress_post_id')
+                         ->orWhere('events_users.role', '=', Role::HOST);
```
ngm, line 394: I'm not sure how this is working.  It looks like it is saying:

- if the event has been approved, then return it
- or if the event has a host, return it

It doesn't seem to be restricting to the query to any particular host, just saying as long as the event has one, pull it back.  Is that right?  (It might well be!)

I'm not sure what the $by_event flag refers to, so I don't know where this is being called.

app/Party.php
```diff
@@ -394,7 +401,15 @@ public function scopeUpcomingEvents($query, $by_event = false)
 
       return $this->join('groups', 'groups.idgroups', '=', 'events.group')
             ->join('users_groups', 'users_groups.group', '=', 'groups.idgroups')
-            ->whereNotNull('events.wordpress_post_id')
+            ->leftJoin('events_users', function($join) {
+                $join->on('events_users.event', '=', 'events.idevents');
+                $join->where('events_users.user', '=', Auth::User()->id);
```
ngm, line 406: Does this run?  I thought it was Auth::user() (lowercase)

app/Party.php
```diff
@@ -394,7 +401,15 @@ public function scopeUpcomingEvents($query, $by_event = false)
 
       return $this->join('groups', 'groups.idgroups', '=', 'events.group')
             ->join('users_groups', 'users_groups.group', '=', 'groups.idgroups')
-            ->whereNotNull('events.wordpress_post_id')
+            ->leftJoin('events_users', function($join) {
```
ngm, line 404: We're joining to events for a specific user here, but not in the $by_event path above - any reason?

app/Party.php
```diff
@@ -394,7 +401,15 @@ public function scopeUpcomingEvents($query, $by_event = false)
...
+                $join->on('events_users.event', '=', 'events.idevents');
+                $join->where('events_users.user', '=', Auth::User()->id);
+            })
+            ->where(function($query) {
+              $query
+                  ->whereNotNull('events.wordpress_post_id')
+                  ->orWhere('events_users.role', '=', Role::HOST);
```
ngm, line 411: If it's restricted to a host, would this mean that admins / network coordinators wouldn't see it here until it is approved?  (Just checking - perhaps that's fine, and they see it elsewhere.)

ngm (commented):
I think this method (or the controller methods calling it - seems to be primarily PartyController->index?) should have some tests around it.  It's a bit opaque as to what it is doing, and I am not really sure whether making these changes will have an effect on existing functionality.

Ideally we would have tests for what an admin, a network coordinator, a host, and a Restarter see depending on whether an event is approved or not.

app/Party.php
edwh: I remember looking at this, but I can't remember what I thought.  I can't find a use of this by_event flag at all, though I might not have got my head round scoped stuff properly.  Can you see if you can?

app/Party.php
edwh: Hmmm.  It can't be right, will change.

app/Party.php
edwh: Will discuss above.

app/Party.php
edwh: They wouldn't see it unless they'd created it, I don't think, but they'd be seeing it in the list of events to approve.

resources/lang/en/events.php
```diff
@@ -201,4 +201,5 @@
   'search_country_placeholder' => 'Country...',
   'search_start_placeholder' => 'From...',
   'search_end_placeholder' => 'To...',
+  'invite_when_approved' => 'You can only invite volunteers when this even has been approved'
```
ngm, line 204: typo, even should be event

resources/views/events/view.blade.php
```diff
@@ -155,6 +155,8 @@
             $brand->brand_name;
             $expanded_brands[] = $brand;
           }
+
+          $event->approved = $event->wordpress_post_id !== null;
```
ngm, line 159: Note: We should probably have an isApproved method on events.

ngm: (doesn't need doing now, just a note)

app/Party.php
ngm: To confirm, I also can't see anywhere that the upcomingEvents scope is being used and is being passed a value for the by_events flag.

PR #329, opened 2021-06-16
DOT-1649 Make Geocoder more robust

edwh:
We don't quite understand what input causes this problem, but we can tell what the fix is.


ngm (changes requested):
After trying for a while to try and get the geocoder fail and return false during group creation - it appears that group creation is not using it. :sweat_smile: 

Group creation seems to be using an old method - /home/restart/Code/restarters/app/Http/Controllers/GroupController.php:146 

The old method also sets a country based on the country returned from Google's geocoding.

I think we could probably update Geocoder.php to return the country, and then use it for the group create?

ngm (commented):
Looks good and nice to have deprecated the old Fixometer method.

Is it possible to add tests around the create method in GroupController and the postProfileInEdit and postRegister methods in UserController, for both happy and failing geocoding paths?

edwh:
Yes, it is.  Added those now.

ngm (approved):

PR #332, opened 2021-06-18
More tests

edwh:
We're now down at a handful of existing tests that don't work, which are excluded in phpuninit.xml.
* Some are ones which integrate with other components (wiki, Wordpress, Zapier).  These probably aren't worth getting running in CircleCI any time soon.
* Some are ones which relate to quests.  Can you think about whether these will disappear or should be made to work?
* One (ExampleTest) uses RefreshDatabase.  Perhaps the stuff I added in TestCase should use that too?  This one could probably be made to work but doesn't feel too important.

So we might be able to move on to adding new tests.  See what you think.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -14,7 +14,9 @@
             <exclude>./tests/Feature/Admin/Users/WikiLoginTests.php</exclude>
             <exclude>./tests/Feature/Groups/WordpressGroupPushTest.php</exclude>
             <exclude>./tests/Feature/Microtasks/ExampleTest.php</exclude>
+            <exclude>./tests/Feature/Microtasks/FaultcatTest.php</exclude>
             <exclude>./tests/Feature/Microtasks/MisccatTest.php</exclude>
+            <exclude>./tests/Feature/Microtasks/MobifixTest.php</exclude>
             <exclude>./tests/Feature/Zapier/ZapierNetworkTests.php</exclude>
         </testsuite>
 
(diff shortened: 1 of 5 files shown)


ngm:
> Some are ones which integrate with other components (wiki, Wordpress, Zapier).  These probably aren't worth getting running in CircleCI any time soon.

`ZapierNetworkTests.php` should be includable, as they don't interact with Zapier, just check that the API that Zapier would call is producing the correct values.

> Some are ones which relate to quests.  Can you think about whether these will disappear or should be made to work?

These will either be removed or changed to make work following the end of the ACTION project - so continue to exclude for now if they're not currently working.

>  One (ExampleTest) uses RefreshDatabase.  Perhaps the stuff I added in TestCase should use that too?  This one could probably be made to work but doesn't feel too important.

When I first started setting up the tests I found RefreshDatabase was far too slow, as it tears down and rebuilds the DB by running migrations.  Hence why I ended up using only the truncating of the relevant tables before each test.  In theory that is OK when running the tests on CI, but its a pain when running locally.  If RefreshDatabase could be made more performant then we could use it.  However I don't think we need to.  We know that the migrations are run at least once at the beginning of the whole build process, so we'd know via that if the process of building a new DB from migrations had an issue.

> So we might be able to move on to adding new tests. See what you think.

Sounds good to me! - like we said at retro, either around key areas of business functionality, or around new bugs that come in, seems like the best places to be adding new tests.

PR #336, opened 2021-06-23
DOT-1480 500 error when editing an event with added devices

edwh:
The scenario here is:

- Create an event in the past
- Add a device
- Edit the event

The audit logs contain the new `devices_updated_at `information which is an array, e.g. this:

`{"devices_updated_at":{"date":"2021-06-23 13:37:18.545386","timezone_type":3,"timezone":"Europe\/London"}}`

The method we use to render audit logs fails if a value is an array.

I'm not sure whether or not we ought to be stopping the data getting into the audit log in array form?  But such data now exists in the live DB.  We can make it render acceptably by spotting it and encoding it.

### DIFF
--- a/database/migrations/2021_04_06_113426_printcat_ora_setup.php
+++ b/database/migrations/2021_04_06_113426_printcat_ora_setup.php
@@ -13,7 +13,7 @@ class PrintcatOraSetup extends Migration {
      */
     public function up() {
 
-        if (!Schema::hasTable('devices_faults_tablets_ora_opinions')) {
+        if (!Schema::hasTable('devices_faults_printers_ora_opinions')) {
             Schema::create('devices_faults_printers_ora_opinions', function (Blueprint $table) {
                 $table->increments('id');
                 $table->string('id_ords', 16)->index();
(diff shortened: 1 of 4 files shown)


database/migrations/2021_04_06_113426_printcat_ora_setup.php
```diff
@@ -13,7 +13,7 @@ class PrintcatOraSetup extends Migration {
      */
     public function up() {
 
-        if (!Schema::hasTable('devices_faults_printers_ora_opinions')) {
+        if (!Schema::hasTable('devices_faults_tablets_ora_opinions')) {
```
ngm, line 16: As this is the migration for PrintCat, the table it is checking for should be `devices_faults_printers_ora_opinions`.

ngm (approved):
> I'm not sure whether or not we ought to be stopping the data getting into the audit log in array form? 

As a general principle I think it's OK to allow data being stored in the log in array form.  In this case it feels slightly inconsistent as it actually goes in to the DB as a datetime.  Which must be Laravel doing some automatic conversion of Carbon dates to datetimes behind the scenes before persistence.  So it's in the logs as a Carbon array, and the DB as a datetime.  That said, I don't feel its necessary to do anything extra here - it's just the logs for an admin to look at.

PR #339, opened 2021-06-23
DOT-1470 remove maps API key

edwh:
Diffs include the audit fix from 1480.

### DIFF
--- a/database/migrations/2021_04_06_113426_printcat_ora_setup.php
+++ b/database/migrations/2021_04_06_113426_printcat_ora_setup.php
@@ -13,7 +13,7 @@ class PrintcatOraSetup extends Migration {
      */
     public function up() {
 
-        if (!Schema::hasTable('devices_faults_printers_ora_opinions')) {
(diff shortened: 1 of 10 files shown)


database/migrations/2021_04_06_113426_printcat_ora_setup.php
```diff
@@ -13,7 +13,7 @@ class PrintcatOraSetup extends Migration {
      */
     public function up() {
 
-        if (!Schema::hasTable('devices_faults_printers_ora_opinions')) {
+        if (!Schema::hasTable('devices_faults_tablets_ora_opinions')) {
```
ngm, line 16: Not sure if this is from a different PR, but given this is the printcat migration, it should be checking for presence of `devices_faults_printers_ora_opinions`.

ngm (approved):

database/migrations/2021_04_06_113426_printcat_ora_setup.php
edwh: It is.  I won't muddy things by fixing it in here.

PR #340, opened 2021-06-23
DOT-1473 duplicate group creation

edwh:
Return a meaningful error if you try to create a group that already exists.

### DIFF
--- a/resources/lang/en/groups.php
+++ b/resources/lang/en/groups.php
@@ -182,4 +182,7 @@ return array (
   'no_other_nearby_events' => 'There are currently no other upcoming nearby events.',
   'postcode' => 'Group postcode',
   'groups_postcode_small' => '(Optional)',
+  'duplicate' => 'That group name already exists.  If it\'s yours, please go to the Groups page using the menu and edit it.',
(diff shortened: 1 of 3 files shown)


resources/lang/en/groups.php
```diff
@@ -182,4 +182,7 @@
   'no_other_nearby_events' => 'There are currently no other upcoming nearby events.',
   'postcode' => 'Group postcode',
   'groups_postcode_small' => '(Optional)',
+  'duplicate' => 'That group name already exists.  If it\'s yours, please go to the Groups page using the menu and edit it.',
```
ngm, line 185: Can we change the first part to 'That group name (:name) already exists.' so the user sees what they entered.

ngm (approved):
I've left a comment re the message shown to the user, but I'm not too fussed if you prefer not, given the group name doesn't appear to be included in the other error messages.

Every time I see these parts of the code base I am reminded that they need a good refactor.  So it's good to add a test in each time we change something here, as you did.

resources/lang/en/groups.php
edwh: I think it's already visible in the form below the message, isn't it?

edwh: Fair enough.  Added.

PR #341, opened 2021-06-24
DOT-854 sentry

edwh:
This enables Sentry for Laravel and Vue.  

Needed to up the version of the faker package for my composer install to work.


.env.example
```diff
@@ -85,3 +85,6 @@ DRIP_ACCOUNT_ID=
 DRIP_CAMPAIGN_ID=
 
 CALENDAR_HASH=somehash
+
+SENTRY_LARAVEL_DSN=https://50fd2fa440af4bb4a230f40ca8d8cf90@o879179.ingest.sentry.io/5831645
```
ngm, line 89: OK for this to be in .env.example in source?

(I presume yes, as it's also in the JS side that goes to the client, but just checking :) )

ngm: I presume that we don't want this used in local environments?  So as not to report on errors in local development?

resources/js/app.js
```diff
@@ -1550,6 +1552,16 @@ jQuery(document).ready(function () {
     lastScrollPosition = currentScrollPosition
   })
 
+  // Sentry error
+  Sentry.init({
+    Vue,
+    dsn: "https://50fd2fa440af4bb4a230f40ca8d8cf90@o879179.ingest.sentry.io/5831645",
```
ngm, line 1558: Will this report errors in local dev too?

Perhaps that's OK?

ngm (approved):
All looks good to me, and excited to see this set up!

Just some questions around what happens with regards to local environments.

.env.example
edwh: I think you can filter out non-production before reporting - pushed a change for that.

For Vue I'm less sure.  
* You can configure[ inbound filters](https://sentry.io/settings/the-restart-project/projects/restarters/filters/data-filters/) to drop events, and I normally have it set up to ignore localhost because that's how I develop in a pure Vue environment, but that doesn't apply for us yet.
* You can also write a beforeSend method to suppress before reporting, which I've used to filter out crud.  I can't immediately think what we'd use in the JS code to establish that we were in production - can you?

resources/js/app.js
edwh: See above.

.env.example
edwh: The key isn't really sensitive I don't think.  In theory people could poison our Sentry with other errors, but we'd spot that.

ngm: >  I can't immediately think what we'd use in the JS code to establish that we were in production - can you?

I think Laravel Mix provides provisions values from the env file for access in JS, so `process.env.APP_ENV` may be available in JS.  I've not used it though so not certain.

edwh: It isn't, but NODE_ENV is.  That'll handle the normal npm run watch case.  Pushed.

Anything else?

ngm: LGTM :+1:

PR #345, opened 2021-06-28
DOT-1464 Summernote alternative

edwh:


edwh:
Closed in favour of DOT-1492.

PR #347, opened 2021-06-30
DOT-1463 delete group

edwh:
Add client and server function to allow Administrators to delete groups.

This is on top of the as-yet-unreleased DOT-810, for testing reasons.


app/Group.php
```diff
@@ -243,6 +244,27 @@ public function getLocation()
         return rtrim($this->location);
     }
 
+    public function canDelete() {
```
ngm, line 247: Minor one - not for now but for retro - PHP code guidelines that I've seen (PSR-2, Laravel's own) tend to go with opening brace on new line.

app/Http/Controllers/GroupController.php
```diff
@@ -766,22 +790,55 @@ public function edit(Request $request, $id)
...
-    // TODO: is this alive?
     public function delete($id)
     {
-        if (FixometerHelper::hasRole($this->user, 'Administrator')) {
-            $r = $this->Group->delete($id);
+        $group = Group::where('idgroups', $id)->first();
+
+        $name = $group->name;
+
+        if (FixometerHelper::hasRole(Auth::user(), 'Administrator') && $group->canDelete()) {
```
ngm, line 799: Not big deal here, but we have a method on user called hasRole which would be better to use in general.  I feel like FixometerHelper class in general should be deprecated over time.

app/Http/Controllers/GroupController.php
```diff
@@ -766,22 +790,55 @@ public function edit(Request $request, $id)
...
+            $r = $group->delete($id);
+            $r = true;
             if ( ! $r) {
-                $response = 'd:err';
+                return redirect('/user/forbidden');
```
ngm, line 803: Does this route still exist?

resources/js/components/GroupActions.vue
```diff
@@ -67,6 +91,12 @@ export default {
       })
 
       this.$emit('left')
+    },
+    deleteGroup() {
+      this.$refs.confirmDelete.show()
+    },
+    async deleteConfirmed() {
+      window.location = '/group/delete/' + this.idgroups
```
ngm, line 99: Is it OK that the delete action is a GET?  As it is behind permission checks, probably OK.  Possibly a more philosophical point, could discuss at retro.

tests/Feature/Groups/GroupDeleteTest.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Tests\Feature\Groups;
+
+use App\Group;
+use App\Role;
+use Tests\TestCase;
+
+class GroupDeleteTest extends TestCase {
```
ngm, line 9: Should we also have a test for the check regarding not being able to delete if past events / devices exist?

ngm (approved):
LGTM :+1: Some minor comments, no blockers though.  Could add in an extra test if time permits.

app/Group.php
edwh: I've changed my IDE.

app/Http/Controllers/GroupController.php
edwh: Thanks.  Changed.

app/Http/Controllers/GroupController.php
edwh: Yes - see https://restarters.net/user/forbidden

resources/js/components/GroupActions.vue
edwh: No, but our use of HTTP verbs is all over the place.  Added to retro.

tests/Feature/Groups/GroupDeleteTest.php
edwh: Yes, and adding that finds bugs.  Tests do that.  Can you rereview?

ngm (approved):
Great, changes all look good now.

PR #350, opened 2021-07-03
Dat 567 battcat status page

zenlan:
Adds missing file that should properly fix the broken test


zenlan:
Grr, want to merge to dev not master. Should do this after coffee!

PR #352, opened 2021-07-05
DOT-1487 automated tests

edwh:
(Wrong branch name)

Removed some code I think is dead, and added some more tests.

The permissions model enforced by UserController@profile looks wrong to me, but I've added tests for the current behaviour.


tests/Feature/Users/ProfileTest.php
```diff
@@ -0,0 +1,121 @@
...
+        // When logged out should throw an exception.
+        try {
+            $response = $this->get('/profile');
+            $this->assertFalse(TRUE);
+        } catch (AuthenticationException $e) {
+            // Success case.
+        }
+
+        // When logged in should be able to see.
+        // TODO I'm not convinced that viewing /profile is ever reachable, though /profile/id is.
```
ngm, line 30: /profile with no id shows you your own profile, I believe.

tests/Feature/Users/ProfileTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+        // When logged out should throw an exception.
+        try {
+            $response = $this->post('/user/edit/' . $user1->id, $editdata);
+            $this->assertFalse(TRUE);
+        } catch (AuthenticationException $e) {
+            // Success case.
+        }
+
+        // TODO These are the behaviours as coded, but they seem weird.
```
ngm, line 64: Agreed - I think right now, it should be only that a user can act on themselves.  Or an admin can act on anyone.

With a possible future addendum that a network coordinator could work on users in their network.

ngm (approved):

tests/Feature/Users/ProfileTest.php
edwh: But do we link to it anywhere?

tests/Feature/Users/ProfileTest.php
edwh: Will raise bug as agreed.

tests/Feature/Users/ProfileTest.php
ngm: Ah, right - at present, I don't know actually!

The only thing I recall related to this, was allowing for the ability to link from elsewhere, e.g. Talk, or the Wiki, to the user's profile page, without requiring knowing their id in Restarters.

PR #359, opened 2021-07-12
DOT-1493 test Discourse

edwh:
This creates an isolated dev environment for Restarters using Docker Compose,
- Create a Dockerfile for Restarters
- Create a Docker compose including Restarters, a DB, phpMyAdmin, Mailhog
- Change some hardcoded Discourse urls to use env().
- Remove the mysql_testing database environment.
- Add a method to set up Discourse testing environment in TestCase::setDiscourseTestEnvironment.

The config works off .env.example, which is:
- munged in Dockerfile
- overridden in TestCase::setDiscourseTestEnvironment
I'm not sure that's very clean, but I don't like having completely separate .env files for testing because it's an extra thing to keep in sync.


resources/views/battcatora/info-modal.blade.php
```diff
@@ -20,7 +20,7 @@
                 <p>@lang('battcatora.info.body-s3-p1', ['url' => 'https://openrepair.org/about/'])</p>
                 <hr>
                 <h5>@lang('battcatora.info.body-s4-header')</h5>
-                <p>@lang('battcatora.info.body-s4-p1', ['url' => 'https://talk.restarters.net/t/5030'])</p>
+                <p>@lang('battcatora.info.body-s4-p1', ['url' => env('DISCOURSE_URL') . 't/t/5030'])</p>
```
ngm, line 23: This doesn't seem correct - it seems to have an extra 't'.

I think it should be: ['url' => env('DISCOURSE_URL') . '/t/5030'])</

resources/views/printcatora/info-modal.blade.php
```diff
@@ -20,7 +20,7 @@
                 <p>@lang('printcatora.info.body-s3-p1', ['url' => 'https://openrepair.org/about/'])</p>
                 <hr>
                 <h5>@lang('printcatora.info.body-s4-header')</h5>
-                <p>@lang('printcatora.info.body-s4-p1', ['url' => 'https://talk.restarters.net/t/4664'])</p>
+                <p>@lang('printcatora.info.body-s4-p1', ['url' => env('DISCOURSE_URL') . 't/t/4664'])</p>
```
ngm, line 23: As above.

resources/views/tabicatora/info-modal.blade.php
```diff
@@ -20,7 +20,7 @@
                 <p>@lang('tabicatora.info.body-s3-p1', ['url' => 'https://openrepair.org/about/'])</p>
                 <hr>
                 <h5>@lang('tabicatora.info.body-s4-header')</h5>
-                <p>@lang('tabicatora.info.body-s4-p1', ['url' => 'https://talk.restarters.net/t/5030'])</p>
+                <p>@lang('tabicatora.info.body-s4-p1', ['url' => env('DISCOURSE_URL') . 't/t/5030'])</p>
```
ngm, line 23: As above.

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+RUN wget https://getcomposer.org/composer-1.phar
+
+# Point at the DB server
+RUN sed -i 's/DB_HOST=.*$/DB_HOST=restarters_db/g' .env
+
+# Turn off session domain, which causes problems in a Docker environment.
+RUN sed -i 's/SESSION_DOMAIN=.*$/SESSION_DOMAIN=/g' .env
+
+# Change the Discourse host to point at the one defined in docker-compose
+RUN sed -i 's/DISCOURSE_URL=.*$/DISCOURSE_URL=http://restarters_discourse:3000/g' .env
```
ngm, line 40: This gave me an error, as mentioned in Jira.  Perhaps doesn't like the slashes in the url?

phpunit.xml
```diff
@@ -37,10 +37,7 @@
         <env name="SESSION_DRIVER" value="array"/>
         <env name="QUEUE_CONNECTION" value="sync"/>
         <env name="MAIL_DRIVER" value="array"/>
-        <env name="DB_CONNECTION" value="mysql_testing"/>
-        <env name="DB_TEST_DATABASE" value="restarters_db"/>
-        <env name="DB_TEST_USERNAME" value="restarters"/>
-        <env name="DB_TEST_PASSWORD" value="s3cr3t"/>
+        <env name="DB_CONNECTION" value="mysql"/>
```
ngm, line 40: I had this as locally I have a dev DB (to test things while coding them, or just testing another PR) that has close to prod data in it, and a separate testing DB.  

If I run tests locally now this would truncate tables in my dev DB I think.  How do you have this set up on your box?

ngm (changes requested):
Great stuff!

Some things:

- There seems to be an error in docker_run
- Some of the discourse url find/replace don't seem right
- wondering how you are running tests locally following removal of mysql_testing

resources/views/battcatora/info-modal.blade.php
edwh: Yes, I spotted that and it should already be fixed.

resources/views/printcatora/info-modal.blade.php
edwh: Fixed.

Dockerfile
edwh: Fixed.

phpunit.xml
edwh: Ah, ok - I didn't realise it was actively used.  

When I run the tests it does indeed wipe my database, which I live with. So that use of a separate environment doesn't seem to be working for me, which I don't understand.  But I'll reinstate that stuff and then regenerate my Docker environment to check it still works.

docker-compose.yml
```diff
@@ -0,0 +1,204 @@
+# This is a Docker Compose development environment for Restarters.  It:
+# - relies on you setting up www.example.com in your hosts file.  TODO this name could be better but needs some configuration.
+# - a web server on http://www.example.com:8001 which serves up restarters.  Admin user is jane@bloggs.net/passw0rd.
+# - phpMyAdmin on http://www.example.com:8002.  Use host restarters_db, user root/s3cr3t
+# - Discourse on http://www.example.com:8003.  User is someuser/mustbetencharacters; see below about hostname.
+#   TODO This is stock Discourse - no theme.  Would be nice to have the theme, but that doesn't matter so far as
+#   testing the integration.
+# - Mailhog on http://localhost:8025.
+#
+# TODO If your host is Linux, this probably only works if you run as root.
```
ngm, line 10: I got confused a bit by this, suggest change to perhaps: 

>    On the machine where you will be using your browser to access the Restarters / Discourse sites once set up on Docker, add a record to your hosts file to point the domain restarters.docker to the machine where you are installing Docker.

(then use restarters.docker instead of www.example.com in the rest of the docs)

docker-compose.yml
```diff
@@ -0,0 +1,204 @@
...
+#
+# First set www.example.com in your hosts file to whatever your host is - e.g. localhost or a VM.
+#
+# Then start using:
+# - docker-compose up -d
+#
+# The Restarters web server won't be available for a couple of minutes after that finishes - it will be doing a composer
+# install and artisan migrate. Discourse also takes quite a few minutes to finish initialising.
+#
```
ngm, line 19: Suggest adding here:

>  Run docker-compose logs to check for any obvious errors from docker-compose.

docker-compose.yml
```diff
@@ -0,0 +1,204 @@
...
+# Once Discourse is up then log in using someuser/mustbetencharacters, and:
```
ngm, line 20: Suggest change to:

> Once Discourse is up, then:

 - Log in to http://restarters.docker:8003 using using someuser/mustbetencharacters
 - etc etc same as currently there

docker-compose.yml
edwh: I'll add the comment, but I want to stick with www.example.com for now as it's the default configuration in (I think) the Discourse container, and I don't want to need to do another tear-down and retest if possible.

ngm: A-ha, yes no problem - didn't realise it was used internally.

docker-compose.yml
edwh: Will do example change of domain.

ngm (approved):
Great stuff!  All working for me now.

PR #372, opened 2021-07-28
DOT-1503 regression tests

edwh:
Bit more testing improvements.


tests/Feature/Groups/GroupNetworkCreateTest.php
```diff
@@ -22,21 +22,6 @@
 
 class GroupNetworkCreateTest extends TestCase
 {
-    public function setUp()
```
ngm, line 25: Is this not needed for this test anymore?

edwh: I don't this so, no - only the "delete from audits" line was any different from the parent method, so I moved that in there.  I'm trying to reduce the number of different setup methods all doing much the same thing, and I noticed this one.

PR #375, opened 2021-08-02
DOT-1492 summernote alternative

edwh:
The Quill is mightier than the Summernote.  

Also removes some dead code which referenced the `rte` class but which is no longer used.


ngm (changes requested):
Code all looks good in general - some comments in Jira though for a couple of things that need looking at.

PR #376, opened 2021-08-04
DOT-1513 sentry automated tests

edwh:
Fix a couple of sentry bugs and add some more coverage tests.


tests/Feature/Users/UserAdminTest.php
```diff
@@ -0,0 +1,51 @@
...
+            ],
+        ];
+    }
+
+    /**
+     *@dataProvider provider
+     */
+    public function testUsersPage($cansee) {
+        // Fetch the list of all users and check that we're in it.
+        $admin = factory(User::class)->states('Administrator')->create();
```
ngm, line 39: Seems like this test is expecting to check this for the various users, based on the data provider, but only seems to be testing it for an admin?

ngm (commented):
Just a comment on UserAdminTest - is it doing what you expect it to be doing?

tests/Feature/Users/UserAdminTest.php
edwh: You're right.  Fixed.

PR #377, opened 2021-08-05
DOT-1500 non geocodeable

edwh:
Refactor the event creation into Vue.  Use vuelidate to validate, including that the address can be geocoded.


resources/js/components/VenueAddress.vue
```diff
@@ -3,20 +3,23 @@
     <b-col md="7">
       <div class="form-group">
         <label :for="$id('address-autocomplete')">{{ __('events.field_event_venue') }}:</label>
+<!--        TODO Moves on focus-->
```
ngm, line 6: When do we want to do this TODO?

resources/js/components/VenueAddress.vue
```diff
@@ -25,6 +28,7 @@
       </div>
     </b-col>
     <b-col lg="5">
+<!--      TODO Map doesn't show on first load of event edit page.-->
```
ngm, line 31: When do we want to do this TODO/

ngm (approved):
Code all looks good to me, a lot of it changed so I've not followed all of the details.

I note there's still a couple of TODOs - can you log somewhere to do them, if not being done for this story?

Noting to self that the way duplication works has changed - but tested and all worked for me.

Noticed a change in behaviour regarding when a user is in charge of only group.  Will log that in Jira.

PR #378, opened 2021-08-09
DOT-1514 duplicate notifications

edwh:
Test and ensure that we don't send duplicate notifications for event creation.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -336,7 +336,7 @@ class PartyController extends Controller
                     foreach ($party->associatedNetworkCoordinators() as $coordinator) {
                         $usersToNotify->push($coordinator);
                     }
-                    Notification::send($usersToNotify, new AdminModerationEvent([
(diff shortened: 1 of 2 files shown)


ngm (changes requested):
Logic looks good, but doesn't the test (or perhaps a separate test) need to test for when the *same* user is both an admin with the preference and a coordinator of the network?

edwh:
I checked before the fix, and it was sending dups as it stood.

But probably better to make that explicit in the test, so I'll do that.

From: "Neil M" ***@***.***>
To: "TheRestartProject/restarters.net" 
***@***.***>
Cc: "Edward Hibbert" ***@***.***>; "Author" 
***@***.***>
Sent: 09/08/2021 17:30:30
Subject: Re: [TheRestartProject/restarters.net] DOT-1514 duplicate 
notifications (#378)

>
***@***.*** requested changes on this pull request.

ngm (approved):
Great, all looks good with the new test :+1:

PR #380, opened 2021-08-09
DOT-1504 deprecated Discourse call

edwh:
Retire use of a deprecated call.


ngm (commented):
Out of interest, what's the reason for no switching away from using the `/admin/users/list/all.json` endpoint?  It seems like it was marginally simpler and less prone to rate limiting.

edwh:
Judging by discussion I've seen, @ngm It only returns the first 100 users and doesn't support paging.

edwh:
Prematurely merged this PR into develop, so follow-up fixes merged into develop manually.

PR #383, opened 2021-08-16
DOT-1474 delete events

edwh:
This adds a canDelete method for events, exposes it to the client so that we can render the relevant option in the menu, and tests the server behaviour.


ngm (approved):
Code looks good!

A question around the functionality of the delete though - adding to Jira.

PR #388, opened 2021-08-18
DOT-1521 password recovery and reset

edwh:
This adds testing for password recovery and reset, and also triggers the event for a change of password.  This may be responsible for some failed wiki logins.

I tested that this worked, and then changed the hash mechanism to be consistent with the code elsewhere.  That still worked, so presumably the old code produced the same output hash.


ngm (approved):
Looks good, and appreciate the comprehensive test and the refactoring away from $_SERVER.

ngm (changes requested):
Actually after testing I don't think this works in propagating the change through to MediaWiki.

In the ChangeWikiPassword event handler, it seems to rely on the presence of an input called 'new-password' in the request.  Which I don't think exists when the PasswordChanged event is raised in the reset() method.

Seems like a flaw in the ChangeWikiPassword handler - feels like the password should be passed along with the event, not pulled out of the request.

edwh:
Superceded by RES-1521 now.

PR #389, opened 2021-08-18
DOT-1494 more sentry logging

edwh:
Listen on logs and pass errors to Sentry.  

Fixed an example of that which I see in dev.

### DIFF
--- a/app/Providers/OurSentryLogging.php
+++ b/app/Providers/OurSentryLogging.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace App\Providers;
+
+use Illuminate\Log\Events\MessageLogged;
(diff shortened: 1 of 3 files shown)


app/Providers/OurSentryLogging.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * Bootstrap services.
+     *
+     * @return void
+     */
+    public function boot()
+    {
+        Event::listen(MessageLogged::class, function (MessageLogged $e) {
```
ngm, line 28: Ah, that's handy that there's a MessageLogged event :+1:

ngm (approved):

PR #390, opened 2021-08-23
DOT-1116 remove build assets

edwh:
See JIRA for details.


ngm (approved):
All looks good (and SO LOVELY to see over 5000 files removed :tada: )

Other comments discussed in Jira.

PR #395, opened 2021-09-03
DOT-1544 discourse signin problems

edwh:
Add quotes into .env.example round Discourse secret, as that caused problems on the live system when it contained a hash.

This fixes a couple of live issues seen after removing the assets from git:

- GDPR cookie notice JS missing
- IFRAME CSS missing


ngm (approved):
:+1:

PR #396, opened 2021-09-03
DOT-1538 incorrect no location set message

edwh:
Return user location to dashboard page, then pass that down and use it to decide whether to display the message about no location being set.

Add some testing of the nearby groups, and fix an issue where some assertions weren't being tested properly.


edwh:
Sorry, a commit has snuck in from DOT-1529.  It'll all come out in the wash/merge.

ngm (approved):

ngm (approved):

PR #398, opened 2021-09-06
DOT-1536 hide inactive groups

edwh:
Rationalise querying of nearby groups into a single method using a common distance value of 50km as per DOT-1535, then filter out groups tagged as inactive.

Remove a bit of dead code.


tests/Feature/Dashboard/BasicTest.php
```diff
@@ -52,11 +52,11 @@ public function testPageLoads($city, $country, $lat, $lng, $nearbyGroupCount)
                 'see-all-topics-link' => env('DISCOURSE_URL').'/latest',
                 ':is-logged-in' => 'true',
                 'discourse-base-url' => env('DISCOURSE_URL'),
-                ':new-groups' => '0',
             ],
         ]);
 
         $this->assertEquals($nearbyGroupCount, count(json_decode($props[0][':nearby-groups'], TRUE)));
+        $this->assertEquals($nearbyGroupCount, count(json_decode($props[0][':new-groups'], TRUE)));
```
ngm, line 59: Is this right? Expected is looking at $nearbyGroupCount, but actual is looking at new groups.

ngm (approved):
Looks good, just a comment on the test for new groups.

tests/Feature/Dashboard/BasicTest.php
edwh: I think I'm missing your point.  There's a count() call in there?

ngm: It seems to be counting the prop for :new-groups, but checking this against $nearbyGroupCount (rather than say $newGroupCount).  I haven't looked further up though, so maybe that's right!

edwh: Ah, I see.

"New groups" means "upto 3 groups which are nearby and created in the last month".  In this case, as the group has just been created, the "new groups" and "nearby groups" are the same thing.  

I think the idea is that there's no point showing you groups that are new but not nearby, otherwise we'd show people in Scotland new groups in Belgium.

PR #399, opened 2021-09-06
DOT-1534 All all groups link to nearby groups.

edwh:
Add a link to the nearby section, and make the link styles consistent within this section.

### DIFF
--- a/resources/views/user/profile/calendars.blade.php
+++ b/resources/views/user/profile/calendars.blade.php
@@ -43,6 +43,7 @@
 
     <div class="input-group input-group-select2 mb-3">
         <select class="form-control select2-with-input-group" id="inputGroupSelect02">
+        @php($first_option=null)
         @foreach ($all_group_areas as $area)
             @if($loop->first)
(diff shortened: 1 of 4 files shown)


resources/views/user/profile/calendars.blade.php
```diff
@@ -43,14 +43,17 @@
 
     <div class="input-group input-group-select2 mb-3">
         <select class="form-control select2-with-input-group" id="inputGroupSelect02">
+        @php($first_option=null)
```
ngm, line 46: What does this do?

ngm (approved):
The see all groups changes look good - just a question on what the change to the calendar links in a user's profile is doing/fixing.

resources/views/user/profile/calendars.blade.php
edwh: There's a template error that you see if you go to the Profile page on a system which is empty, as my test system sometimes is.  If there are no areas then first_option is not set and causes an error.  I slapped this in to make it load.

PR #400, opened 2021-09-08
DOT-1525 cant remove volunteer

edwh:
Rework removal of volunteer to use the id in events_users rather than a user id which may not exist.


app/Http/Controllers/PartyController.php
```diff
@@ -1005,34 +1005,38 @@ public function updateVolunteerQuantity(Request $request)
...
-                }
+        if ($volunteer) {
+            $event_id = $volunteer->event;
 
-                //Return JSON
-                $return = [
-                    'success' => true,
-                ];
+            // Has current logged-in user got permission to remove volunteer?
+            if ((Fixometer::hasRole(Auth::user(), 'Host') && Fixometer::userHasEditPartyPermission($event_id, Auth::user()->id)) || Fixometer::hasRole(Auth::user(), 'Administrator')) {
```
ngm, line 1021: Whenever I see this kind of thing, makes me think we should have a tech debt ticket for permissions.

resources/js/app.js
```diff
@@ -566,15 +562,12 @@ function initAutocomplete() {
...
-        user_id : user_id,
-        event_id : event_id
+        id : id,
       },
       datatype: 'json',
       success: function(json) {
         if( json.success ){
-          jQuery('.volunteer-' + user_id).fadeOut();
-          jQuery('#'+type+'-counter').text();
-          counter.text( current_count - 1 );
```
ngm, line 577: Is this decremented some other way (or no longer needs decrementing?)

resources/js/store/attendance.js
```diff
@@ -17,12 +17,14 @@ export default {
     set(state, params) {
       Vue.set(state.list, params.idevents, params.attendees)
     },
-    remove(state, params) {
-      let newarr = state.list[params.idevents].filter((a) => {
-        return a.user !== params.userId
-      })
+    remove(state, id) {
+      state.list.forEach((list, idevents) => {
```
ngm, line 21: Why is the forEach needed now?

ngm (commented):
Code looks good, I'll have a local test later.

resources/js/store/attendance.js
edwh: Added comment.

resources/js/app.js
edwh: This is old code which used to manually update counts on the main parent page from within the child modal.  But the main page is now in Vue, so it gets updated because the store changes.

app/Http/Controllers/PartyController.php
edwh: Added to retro.

ngm (approved):

PR #401, opened 2021-09-08
DOT-1535 nearby groups radius

edwh:
1. - Change text for nearby groups.
2. - Show group distance.
3. - Don't limit number of groups as much.


resources/lang/en/groups.php
```diff
@@ -164,7 +164,9 @@
   'quick_access' => 'Groups you create or follow appear below for quick access.',
   'no_groups_mine' => '<b>If you can\'t see any here yet, why not <a href="/group/nearby">follow your nearest group</a> to hear about their upcoming repair events?</b>',
   'no_groups_nearest_no_location' => '<b><p>You do not currently have a town/city set.  You can set one in <a href="/profile/edit">your profile</a>.</p><p>You can also <a href="/group/all">view all groups</a>.</p></b>',
-  'no_groups_nearest_with_location' => '<b><p>It doesn\'t look like there are any groups listed near you yet.</p><p>Would you like to start or add a group?  Have a look at <a href="https://talk.restarters.net/t/how-to-power-up-community-repair-with-restarters-net/1228/2">our resources</a>.</p></b>',
+  'no_groups_nearest_with_location' => '<p>There are no groups within 50 km of your location. You can <a href="/group/all">see all groups here</a>. Or why not start your own? <a href="https://talk.restarters.net/t/how-to-run-a-repair-event-the-restart-party-kit/324">Learn what running your own repair event involves.</a></p>',
+  'nearest_groups'  => 'These the groups that are with 50 km of :location',
```
ngm, line 168: Typos, presumably 'These the groups' -> 'These are the groups' and 'that are with' -> 'that are within'

app/Traits/GlobalScopes.php
```diff
@@ -1,41 +0,0 @@
...
-use DB;
-
-trait GlobalScopes
-{
-    /**
-     * From the start of last month till now.
-     * @author Christopher Kelker 2020-02-18
-     * @return $query
-     */
-    public function scopeCreatedWithinLastMonth($query)
```
ngm, line 15: This no longer needed?

ngm (commented):
Code looks good - will have a local test later.

app/Traits/GlobalScopes.php
edwh: I couldn't find any evidence that it was.  But double-check.

resources/lang/en/groups.php
edwh: Yup.

PR #404, opened 2021-09-20
Adds static emission ratios to .env.example

zenlan:
Refactors FootprintRationCalculator to use static emission ratio
Refactores Device->getWeights() to use static emission ratio
Refactors Stats tests to expect static emission ration value


ngm:
LGTM :+1:   I reran the CircleCI build - all OK now coveralls is not failing.

PR #407, opened 2021-10-11
RES-1570 upcoming events

edwh:
In DashboardController:
- remove some dead code  about $outdated_groups and $inactive_groups.
- reinstate some lost code which returns the list of new groups.
- only include upcoming events and groups where we have joined, not just been invited.

In PartyController:
- remove some variables passed to the blade which are no longer used.
 
In Party:
-  show nearby events for groups where we have been invited but not joined.

In DAshboard/BasicTest, add some more testing of upcoming events.


ngm (approved):
Approved (with a note about some erroneous past events, commented in Jira - but that may be a separate bug)

PR #410, opened 2021-10-11
RES-1555 sentry errors

edwh:
- Change the severity of a few error logs which should be info.
- Allow double-leave of group.
- Fix JS error where google maps API is not included.
- Fix JS error where event has no lat/lng (which shouldn't happen outside testing).  
- Update group invitation code and add a test for it.


ngm (approved):
Looks

PR #411, opened 2021-10-18
RES-1311 can't readd

edwh:
The code used sync(), which doesn't play nice with soft deletes out of the box.  

Added a test to repro and then fix.


ngm (approved):
Code all looks good :+1: 

Kind of surprised that sync doesn't work with soft deletes, but fair dos.

PR #413, opened 2021-10-25
RES-1563 Allow network coordinators to delete events.

edwh:
Rework delete permissions for an event:

- Add a method to query it.  The permissions are different from Edit.  As we know, the permissions could do with being centralised.
- Query that.
- Change/add tests in this area.

### DIFF
--- a/app/Helpers/Geocoder.php
+++ b/app/Helpers/Geocoder.php
@@ -8,9 +8,14 @@ class Geocoder
     {
     }
 
+    private function googleKey() {
+        // We have this so that we can change the key in testing.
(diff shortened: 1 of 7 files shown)


app/Helpers/Geocoder.php
```diff
@@ -8,9 +8,14 @@ public function __construct()
     {
     }
 
+    private function googleKey() {
+        // We have this so that we can change the key in testing.
+        return config('GOOGLE_API_CONSOLE_KEY') ?? env('GOOGLE_API_CONSOLE_KEY');
```
ngm, line 13: TBD in retro.

ngm (approved):

PR #415, opened 2021-10-27
RES-1479 Discourse group for group

edwh:
Add an event handler to create a Discourse group on Restarters group approval.

Add a scheduled command discourse:syncgroups which adds/removes members from the Discourse group.

Text is internationalised using the default language of the network.  If there are multiple networks, there might be multiple copies in different languages.


restart-tech (commented):
Code looks good, just a few minor comments.

I think the only one that needs confirming before approving is the sending of the NewDiscourseMember notification when someone is synced to a network.

PR #419, opened 2021-11-03
RES-1587 too many volunteers

edwh:
The logic for which volunteers are invitable is bugged:

- It included people invited to the group but who have not accepted.  They shouldn't be invitable to an event.
- It included people invited to the event already.  We shouldn't be able to pester people by inviting them twice.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -922,14 +922,15 @@ class PartyController extends Controller
     {
         $group_user_ids = UserGroups::where('group', Party::find($event_id)->group)
         ->where('user', '!=', Auth::user()->id)
+        ->where('status', '=', 1)
         ->pluck('user')
         ->toArray();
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/PartyController.php
```diff
@@ -922,14 +922,15 @@ public function getGroupEmailsWithNames($event_id)
...
         ->toArray();
 
-        // Users already associated with the event.
-        // (Not including those invited but not RSVPed)
+        // Users already confirmed as attending the event.
+        //
+        // We don't want to return users who are already invited - we shouldn't be able to invite twice.
         $event_user_ids = EventsUsers::where('event', $event_id)
         ->where('user', '!=', Auth::user()->id)
-        ->where('status', 'like', '1')
```
ngm, line 932: Good spot.

So this is now saying - give me confirmed group members, minus anyone who has already RSVPed or been invited.

PR #422, opened 2021-11-18
RES-1521 password reset

edwh:
Lots of good stuff.

- Add Mediawiki image to the CircleCI configuration and enable wiki integration.  This means our tests now run with a real wiki.
- Change Mediawiki passwords to use the hashed Laravel password.
- Add a test of password recovery/reset.  This needs the code to be a bit more standard Laravel (avoid $_POST etc).
- Add a command to generate the Mediawiki commands to run from maintenance/ to migrate the passwords.
- Make the test users more unique, because we don't clean out the wiki users between tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,15 @@ jobs:
       # We don't have Discourse on CircleCI.
       - run: sed -i 's/FEATURE__DISCOURSE_INTEGRATION=.*$/FEATURE__DISCOURSE_INTEGRATION=false/g' .env
 
+      # ...but we do have Mediawiki.
+      - run: sed -i 's/FEATURE__WIKI_INTEGRATION=.*$/FEATURE__DISCOURSE_INTEGRATION=true/g' .env
+      - run: sed -i 's/WIKI_URL=.*$/WIKI_URL=http://mediawiki:8080/g' .env
+      - run: sed -i 's/WIKI_DB=.*$/WIKI_DB=bitnami_mediawiki/g' .env
+      - run: sed -i 's/WIKI_USER=.*$/WIKI_USER=user/g' .env
+      - run: sed -i 's/WIKI_PASSWORD=.*$/WIKI_PASSWORD=bitnami123/g' .env
+      - run: sed -i 's/WIKI_APIUSER=.*$/WIKI_APIUSER=user/g' .env
+      - run: sed -i 's/WIKI_APIPASSWORD=.*$/WIKI_APIPASSWORD=bitnami123/g' .env
+
       # Playwright needs the debug bar not to appear
       - run: sed -i 's/APP_DEBUG=.*$/APP_DEBUG=FALSE/g' .env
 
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -28,6 +28,26 @@ jobs:
       - image: mcr.microsoft.com/playwright:focal
         environment:
           NODE_ENV: development
+      - image: 'bitnami/mariadb:latest'
```
restart-tech, line 31: Any reason mariadb over mysql?  Fine by me - just thinking we're on mysql on prod, so...

app/Http/Controllers/UserController.php
```diff
@@ -490,52 +491,39 @@ public function reset()
...
+                $response['danger'] = __('passwords.match');
             } else {
-                $user = $User->where('recovery', '=', $recovery)->first();
-                if (! empty($user)) {
-                    $data = [
+                $email = $user->email;
+                $oldPassword = $user->password;
+
+                $update = $user->update([
                     'password' => crypt($pwd, '$1$'.strrev(md5(env('APP_KEY')))),
```
restart-tech, line 509: Does this need to change to use the Laravel Hash helper?  I thought we changed it already recently - perhaps not.

tests/Feature/Users/PasswordResetTest.php
```diff
@@ -0,0 +1,85 @@
...
+        $response->assertSeeText(__('passwords.sent'));
+
+        Notification::assertSentTo(
+            [$restarter], ResetPassword::class
+        );
+
+        $restarter->refresh();
+
+        // Invalid code
+        $response = $this->post('/user/reset', [
```
restart-tech, line 50: There is no assertion on $response before it is given a new value on line 57 - maybe these lines are redundant?

tests/Feature/Users/PasswordResetTest.php
```diff
@@ -0,0 +1,85 @@
...
+        // Valid but mismatch passwords.
+        $response = $this->post('/user/reset', [
+            'recovery' => $restarter->recovery,
+            'password' => 'newpass',
+            'confirm_password' => 'mismatch'
+        ]);
+
+        $response->assertSeeText(__('passwords.match'));
+
+        // Valid - should redirect to login patch and dispatch password changed event to update the wiki.
```
restart-tech, line 74: typo, patch = path?

restart-tech (commented):
All looks good to me in general - should probably wait until having tested against real wiki before approving.

ngm (commented):
All looks good, although one question around the use of the old password hashing method.

I think we should be updating it to Hash::make, like we were doing here? https://github.com/TheRestartProject/restarters.net/pull/388/files#diff-8e91484f56a38d906339650a2e3af6dbda84b6a43e6d889b792bd55b53a54e76L505

app/Http/Controllers/UserController.php
edwh: Perhaps yes.  Fix pushed.

PR #428, opened 2021-12-15
RES-1591 translate notifications

edwh:
Start work on translating notifications.  This needs more work but this stuff can be released.

ModerationEventPhotosNotificationTest shows the method of getting the mail data from the notification, and tests that the internationalisation process is working by running it on users with different locales. I did AdminModerationEventPhotos first as a proof of concept, though later switched to translating notifications for Restarters.  

The automatic selection of locale, using User::preferredLocale, doesn't seem to be working, so we're passing in the locale explicitly in the notification code.  I've left that code in there though, as a safety net for setting the locale correctly.  If you can see why it doesn't work, let me know - you can test that in ModerationEventPhotosNotificationTest.

The data passed to notifications is a bit inconsistent.  I feel it should pass around the relevant objects (or ids?) rather than extracting the data, and let the notifications themselves work out what to include.  I started doing this in EventConfirmed and GroupConfirmed but decided it was too much work for now.  Follow-on story?

The toArray method seems inconsistently used.  I'm not sure whether it is relevant - I don't think the current client receives notifications in this way.  Should we remove it everywhere?

Other stuff:
- Make the language field for the user longer so that we can handle fr-BE etc.
- Changed all uses of the greeting to be internationalised - search and replace.
- GroupConfirmed notification doesn't seem to be generated but should be; add it in and translate it.
- Removed some dead code (e.g. PartyController::sendCreationNotificationEmail, AccountCreated)
- Translate a bit of text when editing events.

This also includes a fix for the raw HTML problem in email notifications, which is a consequence of a change to markdown handling - see https://github.com/laravel/framework/issues/31065


ngm:
> The data passed to notifications is a bit inconsistent. I feel it should pass around the relevant objects (or ids?) rather than extracting the data, and let the notifications themselves work out what to include. I started doing this in EventConfirmed and GroupConfirmed but decided it was too much work for now. Follow-on story?

Interesting one - thinking about it from MVC perspective, I guess if you treat a notification in the same way as a view, i.e. just a dumb renderer, then you'd say that it's the job of the controller to figure out the correct data, and to pass that to the notification, which just displays it, and should have minimal logic.   Can't say I have really strong feelings one way or another though - let's chat at retro and then make a follow-on story if desired?

> The automatic selection of locale, using User::preferredLocale, doesn't seem to be working, so we're passing in the locale explicitly in the notification code. I've left that code in there though, as a safety net for setting the locale correctly. If you can see why it doesn't work, let me know - you can test that in ModerationEventPhotosNotificationTest.

I've no clue at present how that works sadly.

> The toArray method seems inconsistently used. I'm not sure whether it is relevant - I don't think the current client receives notifications in this way. Should we remove it everywhere?

If memory serves then toArray is used for the 'database' delivery channel, i.e. what ultimately ends up as the in-app notifications.  So it should be used in *some* cases, for sure.  I would say that arguably it should be used for *all* notifications, and that if it isn't being used in some cases then it is a mistake.  Although there could be some arcane reason why it isn't in some cases.  I'd say a retro chat plus follow-on story makes sense for this one too.

ngm (approved):
Code looks good to me - given this has the fix to the footer, I suggest we get this released, revisit some questions raised in follow-on stories, and then finish off the other notifications in follow-on stories.

PR #429, opened 2021-12-20
RES-1607 Wordpress event time

edwh:
Ensure that the event start/end pushed to Wordpress doesn't contain seconds.


ngm (approved):
Looks good to me - so we get the first five chars of start and end time, which I presume then is HH:MM.

I wonder what caused this change in the first place - perhaps the switch to Laravel 6?

ngm (changes requested):
Thinking about it, I think this will need to also be amended in EditWordpressPostForEvent.

edwh:
You're right - added that and a test.

PR #430, opened 2021-12-20
RES-1605 Allow hosts to delete their erroneous events.

edwh:
Change the permissions model a bit, and test.

### DIFF
--- a/app/Helpers/Fixometer.php
+++ b/app/Helpers/Fixometer.php
@@ -151,7 +151,7 @@ class Fixometer
 
         $group = Party::find($partyId)->theGroup;
 
-        if (self::hasRole($user, 'Host') && self::userIsHostOfGroup($partyId, $userId)) {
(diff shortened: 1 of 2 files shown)


app/Helpers/Fixometer.php
```diff
@@ -149,8 +149,13 @@ public static function userHasDeletePartyPermission($partyId, $userId = null)
             return true;
         }
 
+        $group = Party::find($partyId)->theGroup;
+
+        if (self::hasRole($user, 'Host') && self::userIsHostOfGroup($partyId, $userId)) {
```
ngm, line 154: Shouldn't this be passed the group id (not the party id)?

ngm (changes requested):

app/Helpers/Fixometer.php
edwh: Yes, absolutely.  Fantastically this works in the test because both the party and group id are 1 because it's a clean DB.

ngm (approved):
Great, all looks good to me now

PR #432, opened 2021-12-21
RES-1609 refactor event queries

edwh:
We should be moving towards using scopes rather than chains of Eloquent methods to make the code more readable.  And certainly not using raw queries.  

This PR introduces a sensible set of scopes into Party, and starts to use them.  Scopes can and should build on each other.  The performance of that might not be ideal, but for our size of database it's OK, and it means that the key logic exists in just one place, which makes the code a lot more maintainable.

There's plenty more to do here - probably the next step would be to introduce a set of scopes into Group.  But this is a reasonable chunk.


app/Http/Controllers/GroupController.php
```diff
@@ -1108,13 +1108,13 @@ public function volunteersNearby($groupid)
         }
 
         //Event tabs
-        $upcoming_events = Party::upcomingEvents()
-            ->where('events.group', $group->idgroups)
+        $upcoming_events = Party::futureForUser()
+            ->forGroup($group->idgroups)
```
ngm, line 1115: This seems like an interesting combo - future for user for a specific group - is that right?

app/Http/Controllers/PartyController.php
```diff
@@ -153,49 +147,6 @@ public function index($group_id = null)
         ]);
     }
 
-    public function allPast()
```
ngm, line 156: Are these just dead views now?  (allPast and allUpcoming)

app/Party.php
```diff
@@ -381,38 +381,173 @@ public function findLatest($limit = 10)
...
-            ->orderBy('event_date', 'ASC');
+    // Scopes.  Each scope should build on a previous scope, getting more specific as we go down this file.  That
+    // isolates query logic more clearly.
+    private function defaultUserIds(&$userids) {
+        if (!$userids) {
+            $userids = [ Auth::user()->id ];
+        }
+    }
+
+    public function scopeUndeleted($query) {
```
ngm, line 395: I thought Eloquent scoped to undeleted by default?  

That said, it can't hurt for us to make it explicit.  But just thinking that even without this scope, we would still be returned only undeleted, unless we explicitly asked for 'withTrashed()'.

app/Party.php
```diff
@@ -381,38 +381,173 @@ public function findLatest($limit = 10)
...
+        $date = date('Y-m-d');
+        $time = date('H:i');
+        $query = $query->undeleted();
+        $query = $query->whereDate('event_date', '=', "'$date'")
+            ->where('start', '<=', "'$time'")
+            ->where('end', '>=', "'$time'");
+        return $query;
+    }
+
+    public function scopeApproved($query) {
```
ngm, line 448: Woohoo!  This will be nice to have.

app/Party.php
```diff
@@ -381,38 +381,173 @@ public function findLatest($limit = 10)
...
+    public function scopeAttendingOrAttended($query, $userids = null) {
+        // Events this user has attending/is attending.
+        $this->defaultUserIds($userids);
+        $query = $query->undeleted();
+        $query = $query->join('events_users AS aoa', function ($join) use ($userids) {
+            $join->on('aoa.event', '=', 'events.idevents');
+            $join->whereIn('aoa.user', $userids);
+
+            // Check the status so that we exclude any events we have been invited to but not confirmed.
+            $join->where('aoa.status', 1);
```
ngm, line 477: I have a vague memory of a bug once, that because `status` is a string, this specific where statement will evaluate to true when the status field is an invite string that starts with 1.  Would be worth a test to check that.

From memory we had to use something like `where('aoa.status', 'like', '1')` to resolve.

app/Party.php
```diff
@@ -445,94 +580,11 @@ public function scopeUpcomingEventsInUserArea($query, $user)
...
-            ->where(function ($query) use ($user_ids) {
-                $query->whereIn('users_groups.user', $user_ids)
-                    ->orWhereIn('events_users.user', $user_ids);
-            })
-
-            ->select('events.*')
-            ->groupBy('idevents')
-            ->orderBy('events.event_date', 'DESC');
+        $query = $query->future();
+        $query = $query->whereNull('wordpress_post_id');
```
ngm, line 586: Wonder could this use the new 'approved' scope?  Can scopes be magically negated?  e.g. $query->future()->not()->approved() or somesuch?

ngm (commented):
All looks good generally, should be much more readable with this refactor.

Some comments left - mostly just thinking out loud - the only one that needs a proper check is with regards to the status field and equality to 1.

app/Party.php
edwh: Hmmm.  I've definitely fixed some bugs where deleted records were being returned.  But it's possible that this was a soft-deleted row in a join table, which I think is not excluded by default.  Because of that kind of thing I've made it explicit.

app/Party.php
edwh: Thank you.  There is a comment to that effect in PartyController around line 571, so I will duplicate that here.

app/Party.php
edwh: It looks from https://stackoverflow.com/questions/34645129/laravel-negate-local-scope as though there isn't magic to do this.

PR #433, opened 2022-01-04
RES-1608 discourse group logo

edwh:
Enhance Discourse sync to apply any group logo to the Discourse group as a flair URL.  

I’ve not added an automated test, as:

- It’s so dependent on the Discourse API that mocking would make it a bit pointless.
- Testing it properly in the Docker Discourse environment would be quite a lot of work.
- I’ve added logs in the error cases so that they will show up in Sentry.

### DIFF
--- a/app/Services/DiscourseService.php
+++ b/app/Services/DiscourseService.php
@@ -230,30 +230,65 @@ class DiscourseService
                 Log::debug("Sync members for Restarters group $restartId, {$group->discourse_group}, Discourse group $discourseId");
 
                 // Check if the flair_url needs updating.  This keeps the logo in sync with changes on Restarters.
-                Log::debug("Check flair url {$discourseResult['group']['flair_url']} vs {$group->groupImagePath()}");
+                Log::debug("Check Discourse logo {$group->discourse_logo} vs {$group->groupimage->idimages}");
+                if (!$group->discourse_logo || $group->discourse_logo != $group->groupimage->idimages) {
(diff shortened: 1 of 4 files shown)


app/Services/DiscourseService.php
```diff
@@ -230,30 +230,72 @@ public function syncUsersToGroups($idgroups = NULL) {
...
+
+                            if ($response->getStatusCode() === 200) {
+                                // Update the group to record that we've uploaded, then we'll skip this next time
+                                // through.
+                                Log::debug("Updated flair url OK for {$discourseId} to {$group->groupimage->idimages}");
+                                $group->discourse_logo = $group->groupimage->image->idimages;
+                                $group->save();
+                                Log::debug("...saved update to group");
+                            } else {
+                                Log::error("Failed to update flair url for {$discourseId}");
```
ngm, line 291: Wondering if we'll end up with lots of errors in Sentry if this fails - I can't remember if we found a way to log argument values in Sentry outside of the title?

ngm (approved):
Code looks all good to me, though with a note as per Jira that somehow the flair image seems to be set OK on live Discourse.

Minor comment on the way in which the error is logged re: Sentry.

app/Services/DiscourseService.php
edwh: As per JIRA I think at least some flair images have been set up manually.

Good point about Sentry..  I'll remove the ID for now.

You can add "context" to Sentry events, but it's a bit more of a faff.  Generally that time is better spent fixing them :-).

PR #440, opened 2022-01-12
RES-1615 translate more notifications

edwh:
Translate and test more notifications.


app/Http/Controllers/PartyController.php
```diff
@@ -777,7 +777,7 @@ public function notifyHostsOfRsvp($user_event, $event_id)
             $hosts = null;
         }
 
-        if (! is_null($hosts)) {
+        if (count($hosts)) {
```
ngm, line 780: I'm not sure how Laravel in a production environment if this is null - is it OK?  

Recent PHPs will give a warning if reporting is E_ALL.  https://3v4l.org/bZCgq
Laravel doesn't like warnings in production.

app/Notifications/JoinGroup.php
```diff
@@ -74,19 +83,23 @@ public function toMail($notifiable)
         } else { // users not yet on the platform
```
ngm, line 83: This one is an interesting one - it is being sent to users not currently on the platform.  So not sure what `notifiable->language` will be.  We should probably create a separate ticket for this specific notification to think about how to deal with this (and the one for inviting to events, where presumably the same could happen).  We could maybe take the locale from the network the group is in (but then, what if it's in two networks etc etc...)

app/Notifications/ReviewNotes.php
```diff
@@ -52,12 +52,16 @@ public function via($notifiable)
     public function toMail($notifiable)
     {
         if ($notifiable !== null && $notifiable->invites == 1) {
+            $locale = $notifiable->language;
+
             return (new MailMessage)
                   ->subject('A repair has been marked as suitable for the wiki')
```
ngm, line 58: A couple of lines here don't seem to be going through translations.

ngm (changes requested):
Looks great, just:

- a question about whether removing null check will be OK on prod
- a couple of places to create follow up tickets (I'll do these)
- a couple of lines in one of the notifications that don't seem to be made translateable

app/Http/Controllers/PartyController.php
edwh: Thanks - will make it robust against that.

app/Notifications/ReviewNotes.php
edwh: ReviewNotes is only sent to admins, though, right?  The aim for this story is only to have done all the non-admin ones.

ngm: Oh yes, that's fine then.

ngm (approved):

app/Notifications/ReviewNotes.php
edwh: Actually it seems to have been removed in another branch; I think James decided he didn't need it.

app/Notifications/JoinGroup.php
ngm: Created https://therestartproject.atlassian.net/browse/RES-1625

app/Notifications/NotifyHostRSVPInvitesMade.php
```diff
@@ -46,24 +46,28 @@ public function via($notifiable)
      */
     public function toMail($notifiable)
```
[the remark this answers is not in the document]
ngm: Created https://therestartproject.atlassian.net/browse/RES-1626 for us to look at it.

PR #441, opened 2022-01-12
RES-1619 Can't load groups page where user and group location are the same.

edwh:
Distance calculation returns NAN if the user and group are at the same location.

### DIFF
--- a/app/Http/Controllers/GroupController.php
+++ b/app/Http/Controllers/GroupController.php
@@ -790,7 +790,12 @@ class GroupController extends Controller
                 $userlng = Auth::user()->longitude;
 
                 if ($grouplat !== null && $grouplng !== null && $userlat !== null && $userlng !== null) {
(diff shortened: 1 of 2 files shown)


ngm (approved):
Fix looks good.

Admit I am still bamboozled by that magic distance formula, and therefore which part of it fails if the things have the same lat/long.  But just adding a guard against that condition is a good way of fixing the bug without having to dissect/redo the formula.

ngm (approved):
Weird, could have sworn I reviewed this somewhere but can't see my comment...

Looks good, I'm generally bamboozled by that magic distance calculation, so therefore unsure which bit of it would have been failing when the things being compared have identical location.  

But just adding a guard against that case seems a good way of preventing it without having to unravel/redo the whole formula. :+1:

ngm:
OK somehow I reviewed this twice, pleased I said roughly the same thing each time :D

PR #442, opened 2022-01-19
RES-1621 EventCreated notification sent to wrong place.

edwh:
We were passing an EventsUsers instance into Notification::send.  I think this is working in test because the database is clean and all the ids are 1 - added something to the retro.  Changed to pass the user.

### DIFF
--- a/app/Party.php
+++ b/app/Party.php
@@ -1055,7 +1055,8 @@ class Party extends Model implements Auditable
         }
 
         // Notify the person who created it that it has now been approved.
-        $host = User::find(EventsUsers::where('event', $this->idevents)->first());
+        $eu = EventsUsers::where('event', $this->idevents)->first();
+        $host = User::find($eu->user);
 
         if ($host) {
             Notification::send($host, new EventConfirmed($this));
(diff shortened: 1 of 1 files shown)


app/Party.php
```diff
@@ -1055,10 +1055,14 @@ public function approve()
         }
 
         // Notify the person who created it that it has now been approved.
-        $host = User::find(EventsUsers::where('event', $this->idevents)->first());
+        $eu = EventsUsers::where('event', $this->idevents)->first();
```
ngm, line 1058: Are we certain this is always the person who created the event?  

I guess it is relying on:

- the person who created the event always having the lowest id in events_users
- the query always returning ordered by events users id ascending

I think the first one is fair for now, though we might be better off having an explicit 'creator' flag of some kind when we revisit events_users statuses.

The second one - no clue, would be good to have a quick check in Laravel docs.

ngm (commented):
Would be good to clarify the comments before merging.

app/Party.php
edwh: I've added an orderBy which will do until we add a creator.

ngm (approved):

PR #443, opened 2022-01-24
RES-1622 sentry

edwh:
Various:

- Fix Sentry errors in joining event by checking data more carefully.
- Remove some dead postEdit code (please check)
- Add some tests for coverage and internationalise a bit while we're in there.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -682,27 +682,34 @@ class PartyController extends Controller
 
         if (empty($not_in_event)) {
             try {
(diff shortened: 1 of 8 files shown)


ngm (approved):
Great

Couldn't see UserController@postEdit being used in the code anywhere.
Tested all the places in the app that I know of where you can edit a user - they all still worked.

So yes it is indeed dead.

In the process I noticed that resources/views/user/edit.blade.php seems like a dead file too - popped it in the ticket in the log of things to look at next time.

PR #445, opened 2022-01-26
RES-1624 timezones

edwh:
This is the first stage in replumbing for timezone support.  Are you sitting comfortably?

- Timezone fields in event and group.  This is always set in the event at creation, but might be explicit or inherited from the group, which in turn can inherit from network(s).
- New event_start_utc / event_end_utc which will in time replace the event_date/start/end fields.  For now those fields continue for read access via virtual columns and accessors, so that most of the old code can continue unaffected...because we've not changed it yet.
- The intention is that we now never write to those old fields, only the UTC ones.  Mutators added to throw an exception if we try.
- Support both old and new field variations on event create/modify.  This allows the current client to work.
- A bit of faffing with the test environment to get the tests to run ok.
- Set CircleCI timezone to UTC, as we will be doing for our servers.
- A bunch of TODOs added for later.

### DIFF
--- a/app/Group.php
+++ b/app/Group.php
@@ -385,6 +385,7 @@ class Group extends Model implements Auditable
      */
     public function upcomingParties($exclude_parties = [])
     {
+        // TODO Timezones
         $from = date('Y-m-d');
 
         if (! empty($exclude_parties)) {
@@ -409,6 +410,7 @@ class Group extends Model implements Auditable
      */
     public function pastParties($exclude_parties = [])
     {
+        // TODO Timezones
         if (! empty($exclude_parties)) {
             return $this->parties()
                 ->where('event_date', '<', date('Y-m-d'))
@@ -450,6 +452,7 @@ class Group extends Model implements Auditable
 
     public function getNextUpcomingEvent()
     {
+        // TODO Timezones
         $event = $this->parties()
             ->whereNotNull('wordpress_post_id')
             ->whereDate('event_date', '>=', date('Y-m-d'))
@@ -693,7 +696,10 @@ class Group extends Model implements Auditable
 
         if (!$timezone) {
             // This should not occur if the networks are set up correctly.
-            throw new \Exception('Group cannot resolve timezone');
+            // TODO Later on we should throw an exception, but only once this code has gone live, as we rely on
+            // the default behaviour during migration.
+            //throw new \Exception("Group {$this->idgroups} cannot resolve timezone");
+            $timezone = 'Europe/London';
         }
 
         return $timezone;
--- a/app/Group.php
+++ b/app/Group.php
@@ -422,7 +422,7 @@ class Group extends Model implements Auditable
                 ->get();
         }
 
-        return $this->parties()->where('event_start_utc', '<', $now)->get();
+        return $this->parties()->where('event_end_utc', '<', $now)->get();
     }
 
     /**
@@ -690,7 +690,7 @@ class Group extends Model implements Auditable
                         // This should not occur if the networks are set up correctly.
                         \Sentry\captureMessage("Problem getting timezone for group {$this->idgroups} - networks conflict with $timezone and {$network->timezone}.  Will use $timezone.");
                         // TODO Convert to exception once groups have timezones set by Neil.
-                        // throw new \Exception("Group is in networks with conflicting timezones");
+                        // throw new \Exception("Group does not have own timezone and is in networks with conflicting timezones");
                     }
                 } else {
                     // First timezone found.
--- a/app/Group.php
+++ b/app/Group.php
@@ -452,11 +452,10 @@ class Group extends Model implements Auditable
 
     public function getNextUpcomingEvent()
     {
-        // TODO Timezones
         $event = $this->parties()
             ->whereNotNull('wordpress_post_id')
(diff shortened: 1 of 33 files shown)


app/Http/Controllers/CalendarEventsController.php
```diff
@@ -130,14 +130,12 @@ public function exportCalendar($events)
             if (! is_null($event->event_date) && $event->event_date != '0000-00-00') {
                 $ical[] = 'BEGIN:VEVENT';
 
-                // Timezone currently fixed to Europe/London, but in future when we
-                // have better timezone support in the app this will need amending.
                 $ical[] = 'TZID:Europe/London';
```
ngm, line 133: This should be $event->timezone?

app/Http/Controllers/PartyController.php
```diff
@@ -123,9 +125,14 @@ public function index($group_id = null)
             }
 
             // ...and any other upcoming events
+            DB::enableQueryLog();
```
ngm, line 128: Needs removing before merge?

app/Http/Controllers/PartyController.php
```diff
@@ -123,9 +125,14 @@ public function index($group_id = null)
...
+
             $other_upcoming_events = Party::future()->
                 whereNotIn('idevents', \Illuminate\Support\Arr::pluck($events, 'idevents'))->get();
 
+            $queries = DB::getQueryLog();
```
ngm, line 133: Needs removing before merge?

app/Party.php
```diff
@@ -396,52 +381,29 @@ public function scopeUndeleted($query) {
...
         // An active event is an event which has started and not yet finished.
-        $date = date('Y-m-d');
-        $time = date('H:i');
         $query = $query->undeleted();
-        $query = $query->whereDate('event_date', '=', $date)
-            ->where('start', '<=', $time)
-            ->where('end', '>=', $time);
+        $now = date('Y-m-d H:i:s');
+        $query = $query->where('event_start_utc', '<=', $now)
+            ->where('event_end_utc', '<=', $now);
```
ngm, line 406: Should this be event_end_utc >= $now?

app/Party.php
```diff
@@ -677,22 +634,10 @@ public function isUpcoming()
...
-        $event_end = new \DateTime($this->event_date.' '.$this->end);
+        $start = new Carbon($this->event_start_utc);
 
-        if ($date_now > $event_end) {
-            return false;
+        if (!$this->$this->isInProgress() && !$this->hasFinished() && $start->isCurrentDay()) {
+            return true;
         }
 
         return true;
```
ngm, line 643: Should return false on final line here?  (Otherwise this method seems to always return true.)

ngm (changes requested):
Fantastic piece of work!

- Wow, completely forgot we had a `SyncEvents` command.  Useful, but I wonder if its working.
- Some changes rely on server time being UTC - we should plan when we're doing this.
- When do we need to set timezones on groups in the default network?
- Commented in a few places where filtering logic appeared not correct.

app/Group.php
```diff
@@ -665,4 +667,41 @@ public function createDiscourseGroup() {
...
+        $timezone = null;
+
+        foreach ($this->networks as $network) {
+            if ($network->timezone) {
+                if ($timezone) {
+                    if ($timezone != $network->timezone) {
+                        // This should not occur if the networks are set up correctly.
+                        \Sentry\captureMessage("Problem getting timezone for group {$this->idgroups} - networks conflict with $timezone and {$network->timezone}.  Will use $timezone.");
+                        // TODO Convert to exception once groups have timezones set by Neil.
+                        // throw new \Exception("Group is in networks with conflicting timezones");
```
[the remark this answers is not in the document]
edwh: Done.

app/Group.php
```diff
@@ -408,14 +408,16 @@ public function upcomingParties($exclude_parties = [])
...
         if (! empty($exclude_parties)) {
             return $this->parties()
-                ->where('event_date', '<', date('Y-m-d'))
+                ->where('event_end_utc', '<', $now)
                 ->whereNotIn('idevents', $exclude_parties)
                 ->get();
         }
 
-        return $this->parties()->where('event_date', '<', date('Y-m-d'))->get();
+        return $this->parties()->where('event_start_utc', '<', $now)->get();
```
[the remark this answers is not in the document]
edwh: Agreed.

app/Http/Controllers/CalendarEventsController.php
edwh: Agreed.

app/Http/Controllers/PartyController.php
edwh: Or indeed now!

app/Party.php
edwh: Yes.

app/Party.php
edwh: Yes.

app/Http/Controllers/PartyController.php
edwh: Yes.

ngm (approved):
Thanks for resolving the comments!  Looks good to me :+1:

PR #446, opened 2022-02-02
RES-1620- Possible fix for 419 login error.

edwh:
This prevents a double-click on the login button, which can result in a 419 page.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -997,18 +997,7 @@ function initAutocomplete() {
     // double-clicking.
     $('#login-form-submit').attr('disabled', 'disabled')
     $('#login-form').submit()
-  })
-
-$('#register-form-submit').on('click', function(e) {
-  e.preventDefault();
-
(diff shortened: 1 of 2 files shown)


resources/js/app.js
```diff
@@ -990,9 +990,14 @@ function initAutocomplete() {
     } else {
       alert('You must consent to the use of your data in order to register');
     }
-
   });
 
+  $('#login-form-submit').on('click', function(e) {
+    // We've seen double submits of the login form, leading to 419 errors.  Prevent the user submitting twice by
+    // double-clicking.
+    $('#login-form-submit').attr('disabled', 'disabled')
```
ngm, line 998: I've noticed in the past that disabled fields don't appear in POST data as a result of being disabled.  Assuming it's not a problem here though, as we won't be checking on the presence of this particular field.

ngm (approved):

PR #447, opened 2022-02-02
RES-1617 Change messageable_level on Discourse groups.

edwh:
Change from "Everyone" to "Only group owners, moderators and admins".  Update the setting in the regular sync.


ngm (approved):
LGTM :+1:

PR #450, opened 2022-02-04
Refactors devices.estimate column from varchar to decimal.

zenlan:
Refactors stats calculations to remove redundant type checking and coalescing.
Refactors edit methods in DeviceController to handle default estimate value.
Removes duplicate estimate assignment in DeviceController edit method.


ngm (approved):
Looks great :+1:   

Nice to be able to remove all of those type and null checks. :tada:

PR #451, opened 2022-02-07
RES-1630 group hosts not owners

edwh:
Rework the syncing to promote/demote hosts on Restarters w.r.t. owners on Discourse.

Rename the sync method as it's now browser than just users.


ngm (approved):
Logic looks sounds to me :+1:

PR #455, opened 2022-02-16
Dat 658 refactor age column

zenlan:
Adds migration and handles input default value


ngm (changes requested):
Having a test, I see that ages display formatted XXX.XX in the UI, which we don't want, e.g. on /fixometer

image

And on an individual event, e.g. http://restarters.test:8000/party/view/2436

image

Can you get it to display how it does currently, i.e. we'll want it left blank nothing shown if the value is 0, otherwise removing any superfluous 0s from either side.

image

ngm (approved):

PR #460, opened 2022-02-23
RES-1647 Event edit broken for timezones.

edwh:
The edit code was missing a key stage in the handling of times, which the create code did correctly.

This is superseded by the later timezone work, but this PR is worth releasing as it beefs up the testing in this area.


ngm (approved):
Ah, I think I see - so for editing we were saving in event timezone, not UTC, but then later we assume it *is* UTC when we convert back to event timezone.  

So when the event timezone doesn't happen to be also UTC, it ends up displaying as a time in the future by the amount of the UTC offset.  Hence why OK for event before BST but not an event that is during BST.

More timezones Inception!

PR #466, opened 2022-03-03
RES-1641 unused API calls

edwh:
Removing some dead API calls and adding a few tests for others.


routes/api.php
```diff
@@ -13,68 +13,39 @@
...
-//     return $request->user();
-// });
-
-Route::get('/homepage_data', function () {
+Route::get('/homepage_data', function () { // Used from DeviceController, tested.
     return App\Http\Controllers\ApiController::homepage_data();
 });
 
-Route::get('/party/{id}/stats', function ($id) {
+Route::get('/party/{id}/stats', function ($id) { // Not used but worth keeping, tested.
```
ngm, line 20: This is used from trp.org
(/var/www/therestartproject.test/wp-content/themes/restart/js/load-specific/fixometer-stats.js)

routes/api.php
```diff
@@ -13,68 +13,39 @@
...
     return App\Http\Controllers\ApiController::partyStats($id);
 });
 
-Route::get('/group/{id}/stats', function ($id) {
+Route::get('/group/{id}/stats', function ($id) { // Not used but worth keeping, tested.
```
ngm, line 24: This is used from trp.org
(/var/www/therestartproject.test/wp-content/themes/restart/js/load-specific/fixometer-stats.js)

ngm (approved):
Great.  All looks good to me.  

Lovely to see so much old code being removed...

routes/api.php
edwh: Ok - added a comment.

routes/api.php
edwh: Added another comment.

PR #467, opened 2022-03-03
RES-1648 discourse blocking

edwh:
Move Discourse access into API calls triggered from the Vue components, rather than retrieving the data in the blades (potentially blocking) and passing down via props.

This adds an extra Vue component when logged in, which affects a bunch of tests in a repetitive way.

Also fix a type error where a Vue prop was defined as Number instead of Array.  I think this only generates a console error in the dev environment, so wouldn't have affected live.


resources/js/components/Notifications.vue
```diff
@@ -0,0 +1,82 @@
...
+      restartersNotifications: 0
+    }
+  },
+  computed: {
+    url () {
+      return this.discourseBaseUrl + '/session/sso?return_path=' + this.discourseBaseUrl + '/u/' + this.discourseUserName + '/notifications'
+    },
+  },
+  async mounted() {
+    const ret = await axios.get('/api/users/notifications/' + this.userId)
```
ngm, line 74: Would `/api/users/<userId>/notifications` be more idiomatic for a restful API?

Recognising that it's an art not a science...

routes/api.php
```diff
@@ -78,3 +78,10 @@
 Route::get('/groups/{group}/events', 'API\GroupController@getEventsForGroup');
 
 Route::get('/devices/{page}/{size}', [App\Http\Controllers\ApiController::class, 'getDevices']);
+
+// Notifications info.  We don't authenticate this, as API keys don't exist for all users.  There's no real privacy
+// issue with exposing the number of outstanding notifications.
+Route::get('/users/notifications/{id?}', 'API\UserController@notifications');
```
ngm, line 84: Is id segment really optional?   We findOrFail in UserController::notifications so seems like we require an id.

Should the route perhaps be `/users/{id}/notifications`, as the id applies to the user resource, not the notification resource.

ngm (commented):
Very nice!  Commented on the pattern of the endpoint, but open to discussion on that.

Other than that all looks great.  Really nice to see a microcosm of the general principle of Vue getting data from an API endpoint.  (I know the components are still tangled up in a top level blade, but still a step in the right direction!)

resources/js/components/Notifications.vue
edwh: Yes, it would.  Changed.

routes/api.php
edwh: Yes.  Changed.

ngm (approved):

PR #472, opened 2022-03-17
RES-1657: bug with calendar end times

ngm:


ngm:
Closing to branch from master instead.

PR #486, opened 2022-04-11
RES-1509 docker compose circle

edwh:
The code changes bely the amount of effort...

This gets Discourse running on CircleCI.  We use a standard Discourse image, and hack an API key into the database.


restart-neil (approved):
Looks great!

PR #489, opened 2022-04-20
RES-1667 Update favicon

edwh:
Update file - and delete an old one which was an invalid file and I think unused.


resources/views/search/index.blade.php
```diff
@@ -77,15 +77,16 @@
                           @foreach($sorted_parties as $groupname => $groupparties)
                           <optgroup label="<?php echo trim($groupname); ?>">
                             <?php foreach($groupparties as $party) { ?>
-                            <option value="<?php echo $party->id; ?>" data-subtext="<?php echo $party->event_date_local; ?>"
+                            <?php $event = App\Party::withTrashed()->find($party->id); ?>
```
restart-neil, line 80: Seems wrong that we'd want to include trashed events in the stats.

That said, that list of events seems to be coming from somewhere further up, which I don't think we want to touch for the purposes of this fix.  The chances of a deleted event having any associated data is also small, so fixing to allow the page to load is the higher priority.  So A-OK.

restart-neil (approved):
All looks good re: the favicon.

Commented regarding the fix for /search, but all good.

PR #490, opened 2022-04-20
RES-1670 Change dropdown to have rotating caret.

edwh:
This:
- Removes the hacky hiding of the dropdown caret using CSS in our code.  Instead use the no-caret option when relevant.
- Add a rotate transition on the caret.

Also:

- Upgrade Bootstrap/Bootstrap Vue versions.  I did this while trying to make it work, which was a red herring, but might as well.
- Remove console log


restart-neil (approved):
Looks good to me

PR #492, opened 2022-04-25
RES-1668 network coord remove

edwh:
Add some extra tests and fix the logic for network coordinators removing from events/groups.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -914,7 +914,7 @@ class PartyController extends Controller
             $event_id = $volunteer->event;
 
             // Has current logged-in user got permission to remove volunteer?
-            if ((Fixometer::hasRole(Auth::user(), 'Host') && Fixometer::userHasEditPartyPermission($event_id, Auth::user()->id)) || Fixometer::hasRole(Auth::user(), 'Administrator')) {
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/PartyController.php
```diff
@@ -914,7 +914,7 @@ public function removeVolunteer(Request $request)
             $event_id = $volunteer->event;
 
             // Has current logged-in user got permission to remove volunteer?
-            if ((Fixometer::hasRole(Auth::user(), 'Host') && Fixometer::userHasEditPartyPermission($event_id, Auth::user()->id)) || Fixometer::hasRole(Auth::user(), 'Administrator')) {
+            if (((Fixometer::hasRole(Auth::user(), 'Host') || Fixometer::hasRole(Auth::user(), 'NetworkCoordinator')) && Fixometer::userHasEditPartyPermission($event_id, Auth::user()->id)) || Fixometer::hasRole(Auth::user(), 'Administrator')) {
```
restart-neil, line 917: Wondering if 

`Fixometer::hasRole(Auth::user(), 'NetworkCoordinator')` 

should be a similar check to the one used in the change in GroupController, i.e. 

`$loggedInUser->isCoordinatorForGroup($group)`

?

So that we're checking if they're a coordinator for this group specifically, not just any network coordinator?

restart-neil (commented):
Just a question on the permissions check in PartyController removeVolunteer.

app/Http/Controllers/PartyController.php
edwh: I think that check will happen inside `userHasEditPartyPermission`.  This is an example of how our permissions need sorting out, obviously.

PR #495, opened 2022-05-04
RES-1690 unable to add devices

edwh:
This fixes a problem we saw on live.

It adds phpunit tests, and also playwright tests (which aren't complete yet).


resources/lang/en/groups.php
```diff
@@ -74,10 +73,8 @@
   'infographic_message' => 'An infographic of an easy-to-understand equivalent of the CO<sub>2</sub> emissions that your group has prevented, such as equivalent number of cars manufactured.',
   'headline_stats_message' => 'This widget shows the headline stats for your group e.g. the number of participants at your parties; the hours volunteered',
   'edit_group' => 'Edit group',
-  'editing' => 'Editing',
   'edit_group_save_changes' => 'Save changes',
   'edit_group_text' => 'Go ahead and change or improve your group\'s profile.',
-  'delete_group' => 'Delete group',
```
restart-tech, line 80: This seems to have been removed, is it supposed to be?

restart-tech (approved):
Looks good - great to have an end-to-end playwright test.

Minor comment about a translation string being removed, just checking in case it was unintentional.

PR #500, opened 2022-05-13
RES-1676 direct email preferences link

edwh:
This rationalises to use internationalisation for the email preferences and adds the URL fragment.  

This fixes a bug in `EventConfirmed`, which was using the wrong URL.


restart-tech (approved):
LGTM :+1:

PR #501, opened 2022-05-13
RES-1685 email notifications

edwh:
Sorry, bad branch name - this is RES-1675.

Tests and fixes the case for this story, by introducing scopes to use rather than raw queries.  Also fixes another similar but less important case.

The various methods in `Group`, and raw queries which access `groups`/`user_groups`, could benefit from refactoring into scopes in the same way we did with `Party` a while back, but I think that's out of scope (pun intended).

### DIFF
--- a/app/Group.php
+++ b/app/Group.php
@@ -646,4 +646,31 @@ class Group extends Model implements Auditable
 
         return $timezone;
     }
+
+    // TODO We've started to refactor into scopes, but this isn't complete yet.
+    public function scopeMembers() {
+        return User::join('users_groups', 'users_groups.user', '=', 'users.id')
+            ->where('users_groups.group', $this->idgroups)
+            ->select('users.*');
+    }
+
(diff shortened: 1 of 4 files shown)


app/Group.php
```diff
@@ -646,4 +646,31 @@ public function getTimezoneAttribute($value)
...
+    }
+
+    public function scopeMembersUndeleted($query) {
+        $query = $query->members();
+        return $query->whereNull('users_groups.deleted_at');
+    }
+
+    public function scopeMembersJoined($query) {
+        $query = $query->membersUndeleted();
+        return $query->where('users_groups.status', 'like', 1);
```
restart-tech, line 664: Perhaps worth commenting here for future travellers why we have to use 'like' rather than equality?

restart-tech (approved):
Looks nice.  Nice to be starting to centralise these member-related queries rather than having them dotted around.

PR #505, opened 2022-05-23
RES-1694 require timezone on Group creation

edwh:
Makes the timezone mandatory on group create (though still optional on edit for backwards compatibility).


restart-tech (changes requested):
According to the acceptance criteria, we want this to be optional on both create and edit.

edwh:
Modified to:
* Have timezone present but optional on create
* Update the timezone on upcoming events when the group timezone is changed.

PR #507, opened 2022-05-30
RES-1696 queuing

edwh:
1. Remove `ShouldQueue` from everything except notifications to admins.
2. Change the default queue to be database.  This will need a .env change on live.
3. Add a test that notifications are queued (and not queued) as we expect.


app/Notifications/BaseNotification.php
```diff
@@ -0,0 +1,23 @@
...
+class BaseNotification extends Notification
+{
+    public function failed($e)
+    {
+        if (gettype($e) == 'string') {
+            \Sentry\captureMessage("Notification failed with $e");
+        } else if ($e instanceof \Exception) {
+            \Sentry\captureException($e);
+        } else {
+            \Sentry\captureMessage('Notification in an unexpected way ' . gettype($e));
```
restart-tech, line 20: Should say 'Notification failed in an unexpected way '?

restart-tech (approved):
Code all looks good (one minor typo)

Will test on dev instance now.

app/Notifications/BaseNotification.php
edwh: Yes - fixed.

ngm (approved):

PR #508, opened 2022-06-01
RES-1495 sentry user errors

edwh:
1. Log to Sentry based on HTTP return code.
2. Log to Sentry for when we use 'danger' and 'warning' (with some exceptions where these errors can legitimately occur).
3. Delete some dead code.
4. Add some tests for coverage.


ngm (approved):
This looks great in terms of Sentry logging user facing errors/warning :+1: 

Some of the old error/warning handling code (and messaging copy) seems pretty clunky.   But perhaps we can resolve if and when we see instances of particular errors actually occurring.

PR #513, opened 2022-06-20
RES-1706 sentry

edwh:
- Fix error with estimate field similar to age field, found in Sentry
- Remove dead code in *Cat controllers.
- Fix a test which runs locally but not on CircleCI, due to device_brand not being nullable in migration.
- Add very basic test of some artisan commands which are scheduled.
- Get Docker compose build working again.


restart-tech (approved):
All looks good - though could you check in with Monique (@zenlan) please to let her know some of the old quest code is being removed, thanks!

PR #521, opened 2022-07-06
RES-1697 email queueing

edwh:
1. Move some duplicate code into BaseNotification.
2. Make BaseNotification queue.
3. Fix up tests.

### DIFF
--- a/app/Notifications/BaseNotification.php
+++ b/app/Notifications/BaseNotification.php
@@ -27,6 +27,7 @@ abstract class BaseNotification extends Notification implements ShouldQueue
      */
     public function via($notifiable)
     {
+        // If user being notified has opted in to receive emails.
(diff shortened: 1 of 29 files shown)


app/Notifications/BaseNotification.php
```diff
@@ -4,12 +4,36 @@
...
+
+    /**
+     * Get the notification's delivery channels.
+     *
+     * @param  mixed  $notifiable
+     * @return array
+     */
+    public function via($notifiable)
+    {
+        if ($notifiable->invites == 1) {
```
ngm, line 30: As I find the mismatch between the column name and what it actually means confusing whenever I see it - and now we only have to do it in one place - could you add a comment here: 

// If user being notified has opted in to receive emails.

ngm (approved):

PR #529, opened 2022-07-19
DAT-702 brand indexes

zenlan:
Adds migration to add indexes on brand columns to devices and brands tables


ngm (approved):
Ran locally, indexes added fine.

PR #531, opened 2022-07-20
RES-1722 suggest product category

edwh:
Allow user to type what they're repairing, and suggest a category from it.


edwh:
Now handled in RES-1787.

PR #547, opened 2022-09-05
RES-1757 don't publish approved events on unapproved groups

edwh:
Both pushing to WordPress and in the API.


ngm (approved):
Code all looks clear - basically whereever we are 'publishing' an event, only do so if the group is approved.

PR #548, opened 2022-09-06
RES-1725 Sentry

edwh:
Fixed some untranslated key issues, which are swamping the Sentry logs so much it’s hard to see anything else.

### DIFF
--- a/app/Helpers/RobustTranslator.php
+++ b/app/Helpers/RobustTranslator.php
@@ -19,9 +19,13 @@ class RobustTranslator extends BaseTranslator
     {
         $translation = parent::get($key, $replace, $locale, $fallback);
 
-        // Check whether this key is something for which we expect a translation.  Exclude things in the JSON files,
-        // i.e. without a dot, as these do contain values which are validly the same as the key in English.
-        if (strpos($key, '.') !== FALSE && $translation === $key) {
+        // Check whether this key is something for which we expect a translation.  Exclude:
+        // - Things in the JSON files, i.e. without a dot, as these do contain values which are validly the
+        //   same as the key in English.
+        // - Audit info, where we often don't expect a translation.
+        if (strpos($key, '.') !== FALSE &&
+            strpos($key, 'event-audits') === FALSE &&
+            $translation === $key) {
             // This is very likely to be an error, where we have failed to translate something or fat-fingered the key.
             \Sentry\captureMessage('Translation not found for ' . $key);
             Log::warning('Translation not found for ' . $key);
--- a/app/Helpers/RobustTranslator.php
+++ b/app/Helpers/RobustTranslator.php
@@ -19,7 +19,9 @@ class RobustTranslator extends BaseTranslator
     {
         $translation = parent::get($key, $replace, $locale, $fallback);
 
-        if ($translation === $key) {
+        // Check whether this key is something for which we expect a translation.  Exclude things in the JSON files,
+        // i.e. without a dot, as these do contain values which are validly the same as the key in English.
+        if (strpos($key, '.') !== FALSE && $translation === $key) {
             // This is very likely to be an error, where we have failed to translate something or fat-fingered the key.
             \Sentry\captureMessage('Translation not found for ' . $key);
             Log::warning('Translation not found for ' . $key);
--- a/app/Listeners/LogInToWiki.php
+++ b/app/Listeners/LogInToWiki.php
@@ -40,8 +40,6 @@ class LogInToWiki
     {
         $user = $event->user;
 
-
-
         if ($user->wiki_sync_status == WikiSyncStatus::CreateAtLogin) {
             Log::info("Need to create " . $user->name);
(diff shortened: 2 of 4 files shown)


app/Group.php
```diff
@@ -464,11 +464,6 @@ public function changesShouldPushToZapier()
         return false;
     }
 
-    public function getMaxUpdatedAtDevicesUpdatedAtAttribute()
```
ngm, line 467: Is this definitely not used?  I remember it being a fairly recent addition.

app/Party.php
```diff
@@ -442,19 +309,6 @@ public function scopeAttendingOrAttended($query, $userids = null) {
         return $query;
     }
 
-    public function scopeInvitedNotConfirmed($query, $userids = null) {
```
ngm, line 445: Definitely not used?  Seems like a recent addition.

app/Party.php
```diff
@@ -968,11 +790,6 @@ public function associatedNetworkCoordinators()
         return $coordinators;
     }
 
-    public function getMaxUpdatedAtDevicesUpdatedAtAttribute()
```
ngm, line 971: Definitely not used?

ngm (commented):
Left a few comments just to double check on a few things that seem fairly recently added but now removed.  (scopes and attributes - as the search term for their usage doesn't necessarily correspond to the actual method name..)

app/Party.php
edwh: Definitely not - searching for InvitedNotConfirmed finds nowt.

app/Group.php
edwh: Nope - search for MaxUpdatedAtDevicesUpdatedAt finds nowt.

app/Party.php
```diff
@@ -467,12 +321,6 @@ public function scopeMemberOfGroup($query, $userids = null) {
         return $query;
     }
 
-    public function scopeHostOfGroup($query, $userids = null) {
```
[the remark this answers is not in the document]
edwh: Nope - search for HostOfGroup shows related methods but no use of this party one.

app/Party.php
edwh: Nope - search for MaxUpdatedAtDevicesUpdatedAt finds nowt.

app/Console/Commands/ImportGroups.php
```diff
@@ -0,0 +1,138 @@
...
+                // Create the group.
+                try {
+                    $group = new \App\Group();
+                    $group->name = $groupname;
+                    $group->location = $location;
+                    $group->latitude = $lat;
+                    $group->longitude = $lng;
+                    $group->country = $country;
+                    $group->website = $website;
+                    $group->free_text = $description;
```
ngm, line 122: In the DB there's a 'shareable_code' field.  Does this get set automatically?  Or do we need to set it here?

app/Console/Commands/ImportMRES.php
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $input = $this->argument('input');
+        $output = $this->argument('output');
+        $networks = $this->option('networks');
```
ngm, line 44: What's the format of this - comma separated list of networks ids?

app/Console/Commands/ImportMRES.php
```diff
@@ -0,0 +1,151 @@
...
+                // Construct.  We don't have a generic description, so it will be empty for most groups.
+                $description = "";
+
+                if ($dates) {
+                    $description .= "<p>Dates et horaires: " . htmlspecialchars($dates) . "</p>";
+                }
+
+                if ($organisateur) {
+                    $description .= "<p>Organisateur: " . htmlspecialchars($organisateur) . "</p>";
+                }
```
ngm, line 131: We also said we would add the Inscription field to the Description.

app/Console/Commands/ImportMRES.php
```diff
@@ -0,0 +1,151 @@
...
+                        [
+                            $groupname,
+                            $location,
+                            'France',
+                            $lat,
+                            $lng,
+                            $website,
+                            $phone,
+                            $networks,
+                            $description,
```
ngm, line 146: We should also be passing 'Ville' as the 'Area' field, somewhere.

ngm (commented):
Some changes needed for RES-1756.

app/Console/Commands/ImportGroups.php
edwh: I think you're right that we need to set it.  Added (in 1756).

app/Console/Commands/ImportMRES.php
edwh: Yes.

app/Console/Commands/ImportMRES.php
edwh: Will add.

PR #551, opened 2022-09-19
RES-1756 Import MRES

edwh:
Add phone number to group.

Review comments from 1725.

### DIFF
--- a/app/Http/Resources/Group.php
+++ b/app/Http/Resources/Group.php
@@ -67,6 +67,16 @@ class Group extends JsonResource
 
     /**
      *     @OA\Property(
+     *          property="phone",
+     *          title="phone",
+     *          description="An optional phone number to contact the group.",
+     *          format="string",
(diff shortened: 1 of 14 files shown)


app/Http/Resources/Group.php
```diff
@@ -65,6 +65,16 @@ class Group extends JsonResource
      *     )
      */
 
+    /**
+     *     @OA\Property(
+     *          property="phone",
+     *          title="phone",
+     *          description="An optional phone number to contact the group.",
+     *          format="string",
+     *          example="https://therestartproject.org"
```
ngm, line 74: The example is wrong.

ngm (approved):

app/Http/Resources/Group.php
edwh: Indeed it is.  Fixed (in develop, post merge).

PR #554, opened 2022-09-19
RES-1770 Firewall invalid emails

edwh:
This checks that emails are valid when we invite to groups/events.

As with many error messages, the error text is not translated.  This isn't great, but I want to get this one in quickly because of the system impact.


ngm (approved):
The change of indentation level on made the diff on PartyController a bit annoying to review, but as far as I can make out, the change is actually not too big - we just guard against invalid emails addresses before going on to do the email sending.  That all looks good to me.  As discussed we need better UX to report back to the user when this happens, but for a future ticket.  Will test locally in the app now.

PR #564, opened 2022-10-07
RES-1760 remove xref

edwh:
1. Add a `logo` column to the `networks` table to store the file path, and migrate the data over.
2. Upload files in NetworkController using `store()`, which is more idiomatic.
3. Put them in `public_uploads/network_logos`, by creating a disk for that in `filesystems.php` and passing a path to `store()`.
4. Change the `logo()` method to generate the appropriate path for Croppa, which will handle generating and caching different sizes as required.
5. Increase logos from 50px to 100px - retina screens means that probably too low res nowadays.
6. Fix a bug in the test, where the regex was greedy and shouldn't be.

### DIFF
--- a/app/Http/Controllers/NetworkController.php
+++ b/app/Http/Controllers/NetworkController.php
@@ -98,7 +98,6 @@ class NetworkController extends Controller
 
             // Store it in the network object.
             if ($path) {
-                error_log("Save logo $path");
                 $network->logo = $path;
                 $network->save();
             } else {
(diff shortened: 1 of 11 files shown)


ngm (approved):
Looks great, much cleaner, and great to have the image scaling too!

Minor comment that I can take or leave, but I noticed the filename when stored with a size is given both a dash and an underscore.

e.g. 

* 8OiZQjyChnqqTsszacvafdxRfJbQyFTFCrC6HeGn.png  becomes
* 8OiZQjyChnqqTsszacvafdxRfJbQyFTFCrC6HeGn-_x100.png

perhaps just needs one?

edwh:
@ngm Thanks.

The format for the files is `-`, then `width`x`height`.  If the width isn't specified, then you'll get `-x`.  But that's as designed.  

See https://github.com/BKWLD/croppa for more.

PR #571, opened 2022-10-19
RES-1780 vueify group event

edwh:


app/Http/Controllers/GroupController.php
```diff
@@ -94,161 +94,16 @@ public function network($id)
...
-                'selectedNetworkId' => $networkId,
-            ]);
-        }
-
-        return view('group.create', [
-            'title' => 'New Group',
-            'gmaps' => true,
-            'selectedNetworkId' => $networkId,
-        ]);
+        return view('group.create');
```
restart-neil, line 106: Is it OK to have done away with the 'selectedNetworkId' here?  Presume it wasn't actually doing anything?

app/Http/Resources/Group.php
```diff
@@ -279,8 +313,10 @@ public function toArray($request)
             'stats' => $stats,
             'updated_at' => Carbon::parse($this->updated_at)->toIso8601String(),
             'location' => new GroupLocation($this),
-            'networks' => $this->resource->networks,
+            'networks' => new NetworkSummaryCollection($this->networks),
+            'tags' => new TagCollection($this->group_tags),
```
restart-neil, line 317: I think here we have an issue to think about, in that some of the data we use internally, is not really relevant (and perhaps shouldn't be seen at all) to our current API users.  For example, at present group tags are only set by admins, yet we're going to be returning them to clients of the API.  Not a big issue for now I don't think, but this general distinction of who sees what via the API would be good to discuss.  We kind of have two use cases - the API for *our client*, and the API to return certain data we want partners to see.  Maybe it's fine for those both to end up being the same thing, but we should discuss.

restart-neil (approved):
- use nginx rather than php artisan serve
- change to role check
- big changes to GroupController
  - in general I like the fact that all of the POST related logic has moved elsewhere (to the API), view and edit logic was always jumbled up in Group/Event controllers
- some things exposed via API now that may only be relevant to our client
- lots of changes to vue files
- we've removed some flash alerts, how do these alerts get displayed now?

app/Http/Controllers/GroupController.php
edwh: Exactly - you can check in the old template.

app/Http/Resources/Group.php
edwh: As discussed at retro.

edwh:
> * we've removed some flash alerts, how do these alerts get displayed now?

The Vue component makes an API call via the store, then sets a flag to trigger display of the text.  See for example use of `failed` in `GroupAddEdit`.

PR #573, opened 2022-10-24
RES-1768 Heroku

edwh:
This is a WIP to set up CircleCI to deploy to Heroku.

1. CircleCI config.yml needs to move to version 2.1, which requires some changes.
2. Add the CircleCI and Heroku config more or less as per https://circleci.com/blog/deploy-laravel-heroku/, though pushing the current Circle branch rather than master.  
3. Heroku only supports PHP 8, which means we need to upgrade some packages.   We need to move to PHP 8 anyway.

Currently:
* Stalled on `hieu-le/wordpress-xmlrpc-client`, which is unmaintained and relies on the deprecated `ext-xmlrpc` PHP extension which isn't available on Heroku.  We should find a replacement for that package.
* We don't yet know how to create a Heroku app per branch.


edwh:
Not progressing.

PR #583, opened 2022-11-09
RES-1786 Improve code coverage

edwh:
Delete some more dead code (please check you agree in case there's something I don't understand about Drip) and add a very very rudimentary test for a bit of the remaining Drip code.


ngm (approved):

ngm:
Agree on the Drip code - we're only using createOrUpdateSubscriber.

PR #585, opened 2022-11-17
RES-1810: Allow client to pass start/end date filters for network's event list

ngm:
Similarly to `app\Http\Controllers\API\GroupController::getEventsForGroupv2`, allow filtering of network events by start and end date of event.

We already had filtering by updated date, but this is for the actual event date.


edwh (approved):
Looks good.

PR #586, opened 2022-11-17
RES 1811 Fix return value for network logo path

ngm:
Since we changed the way in which the image uploads work, we just need to return the string value of the path now.


edwh (changes requested):
The fix looks correct - but shouldn't we have a test?

restart-tech (approved):

PR #596, opened 2022-12-08
RES-1822 wiki collapsible section

edwh:
The wiki CSS has some code which overrides styles in the collapsible section. This seems to be wrong for our purposes - perhaps the overall style has changed and this didn't get updated.

### DIFF
--- a/resources/wiki/css/app.scss
+++ b/resources/wiki/css/app.scss
@@ -134,9 +134,6 @@ body {
     }
 
     div:first-of-type {
-      @extend .d-flex;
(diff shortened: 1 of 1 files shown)


restart-neil (approved):
From eyeballing it, I feel like some of the changes here might change how the main page looks, with its numbering of the collapsible sections.  The rules for `content` that we've removed.

That said, if this both fixes the actually functional issue that Philip experienced, and makes our styles closer to the MediaWiki default styles, then I'm happy.

PR #598, opened 2022-12-19
RES-1633 unapproved events

edwh:
This builds on RES-1101, so the diffs will look wrong until that is merged.

Unapproved events (and devices of those events) should be visible to:
- admins
- network coordinators
- relevant hosts 

So:
1. Add a `userCanSeeEvent` method in `User` to contain this logic.
1. In `ExportController`, add this logic and a new approved column into the export.
2. In `Search`, add this logic.
3. Return the `approved` property in the API for `Party` and `PartySummary`.
4. Remove some debug in `EditWordpressPostForEvent`
5. Fix a bad query in `Party::ofTheseGroups`.
6. In `CalendarEventsController` add this logic and fix up the event status, including for cancelled events.
7. Test this.


app/Http/Controllers/ExportController.php
```diff
@@ -90,32 +94,34 @@ public function devices(Request $request, $idevents = NULL, $idgroups = NULL)
...
-                $co2Diverted = 0;
-
-                if ($device->isFixed()) {
-                    if ($device->deviceCategory->powered) {
-                        $wasteImpact = $device->eWasteDiverted();
-                        $co2Diverted = $device->eCo2Diverted($eEmissionRatio, $displacementFactor);
-                    } else {
-                        $wasteImpact = $device->uWasteDiverted();
-                        $co2Diverted = $device->uCo2Diverted($uEmissionratio, $displacementFactor);
+                if (User::userCanSeeEvent($me, $event)) {
```
ngm, line 97: I think ExportController::devices is called from https://therestartproject.org/download-dataset - a public call with no user as far as I know.  So wonder if this would then mean it returns no data?

edwh: I wouldn't have thought so.  `userCanSeeEvent` handles the case where the user is logged out (i.e. null) and would return true if the event and group have been approved.  But consider.

ngm: Yes perfect, I realised this when I got to the userCanSeeEvent method - if event and group both approved then the event/item data will be visible.

ngm (commented):
In general, it adds a method for checking whether a user can see an event based on event and group approval.  And then amends Calendar, Export, and Search to take this into account.

- Added in a permission check for Calendar Events, which seems good.
- Wondered about ExportController::devices - one place it is called is https://therestartproject.org/download-dataset/ which is a public call - so I wondered if the permission check in userCanSeeEvent is added, whether this would then return no data here.  But as discussed it will still be fine, as it will return devices from events where both the events and groups are approved.
- I get a 'Maximum execution time of 60 seconds exceeded' when trying http://restarters.test:8000/export/devices locally with the new code change - I guess the extra permission check slows it down?  /export/devices completes if I run it off develop.
- Seems correct to filter Search::parties by userCanSeeEvent.
- approved is returned via the API, but events are not filtered dependent on approved.  This is correct, as we let the client side decide what they want to do.

PR #599, opened 2023-01-04
RES-1802 OpenAPI coverage

edwh:
Prototype validation of API responses against OpenAPI schema.


edwh:
Will be handled by RES-1843.

PR #606, opened 2023-01-25
RES-1834 Swagger broken by upgrade

restart-neil:


config/l5-swagger.php
```diff
@@ -1,24 +1,301 @@
...
-        'middleware' => [
-            'api' => [],
-            'asset' => [],
-            'docs' => [],
-            'oauth2_callback' => [],
+    'default' => 'default',
+    'documentations' => [
+        'default' => [
+            'api' => [
+                'title' => 'L5 Swagger UI',
```
restart-neil: @edwh Should be 'Restarters API'?

restart-neil (commented):
Looks great.

Tested locally and all works.

Can't approve it as I created it.

config/l5-swagger.php
restart-neil: Changed it.

PR #607, opened 2023-01-30
RES-1846 Event date is wrong when editing in some timezones

edwh:
The code is using `slice()` rather than getting the date correctly using the timezone.


restart-neil (approved):
Change looks good.

Should we have a test for this?  Presuming it's harder as it's client side though.

edwh:
@restart-neil Yes, it's quite fiddly client-side.  Your call.

restart-neil:
Let's not for this particularly bug, in order to get the fix released.

PR #610, opened 2023-02-01
RES-1822 wiki front page styling

edwh:
This was slightly broken by recent changes:

- Arrow bullets lost
- Header numbers and square lost


restart-tech (approved):
Changes all look fine to me - seems that icons just moved to a different folder, and a couple of changes for the collapsible sections.

PR #618, opened 2023-02-13
RES-1856 Allow "outdated" timezones

edwh:
Some timezones, e.g. Australia/Canberra, are "outdated", and not returned by PHP by default.  We had an inconsistency - we grabbed a list of timezones from MySQL, but then validated it using the PHP list.

It's arguable which we should use, but we should definitely be consistent.  I think we should use the PHP list, because if it doesn't work in MySQL then it should - it's a MySQL config issue.

### DIFF
--- a/app/Http/Controllers/API/GroupController.php
+++ b/app/Http/Controllers/API/GroupController.php
@@ -720,7 +720,7 @@ class GroupController extends Controller
         $longitude = null;
         $country = null;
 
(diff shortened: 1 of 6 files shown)


restart-tech (approved):
So it looks like with this change:

- we now get the list of available timezones from PHP rather than MySQL
- everywhere we do something with timezones, we use ALL_WITH_BC to include outdated ones
- presumably then we expect outdated ones to work fine with MySQL?

PR #619, opened 2023-02-13
RES 1842 reduce N+1 queries

edwh:
These are now usefully spotted by Sentry.  Add some uses of `with()` and caching to reduce database queries.


ngm (approved):
So it looks like everywhere(?) that we query on groups, we now also eagerly fetch the info about networks in.  To avoid that query on the networks running for every group, each time we iterate through a list of groups.

Seems good, I suppose we might to think about it if for example networks themselves ever started eagerly fetching things as well, I guess you can get to a point where you end up pulling the whole DB if you're not careful.

Nice to cache the roles table as well.

PR #621, opened 2023-02-22
RES-1849 Make number of events/groups shown on dashboard have a consistent max

edwh:
Use a max of 5 for groups and upcoming events.


ngm (approved):
Looks good - limiting upcoming events to 5, and increasing number of groups to 5, as per the ticket's request.

PR #623, opened 2023-02-27
RES-1843 Add OpenAPI coverage testing

edwh:
This builds in automatic testing of the OpenAPI specification to our tests.  

We have some cases where we use `anyOf` to return a group/group summary or similar.  For the code to work out which one it is and validate it, we need to add in fields which force the models to be different (not just a subset) and make it required.

Fixes quite a fix issues with the way the API was documented which I hadn't noticed.


ngm (approved):
Looks great to me.

Comprehensive documentation of the schema of all resources that might be returned via the API - and each call links to that schema so you can see what you'd expect back inline from any particular call.

Bravo!

PR #624, opened 2023-02-27
REP-245 Add Platform.sh files

edwh:
Evaluating platform.sh.


edwh:
Closing - we will pursue this via another story in future.

PR #626, opened 2023-03-10
RES-1853 Calculate the number of hours volunteered based on event length

edwh:
This currently includes a script to calculate changes.  We'd probably not release this as it stands, just the small change to the calculation.


ngm (approved):
All looks good :+1:   As you say, for release we'd remove the new/old versions of the calc.

PR #629, opened 2023-03-23
Network Coordinator can't edit device data

edwh:
Permissions code doesn't consider that case.

### DIFF
--- a/app/Helpers/Fixometer.php
+++ b/app/Helpers/Fixometer.php
@@ -203,6 +203,7 @@ class Fixometer
         if (self::hasRole($user, 'Administrator')) {
             return true;
         }
+
         if (self::hasRole($user, 'NetworkCoordinator')) {
             $group = Party::find($eventId)->theGroup;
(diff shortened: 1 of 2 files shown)


app/Helpers/Fixometer.php
```diff
@@ -220,12 +221,22 @@ public static function userHasEditEventsDevicesPermission($partyId, $userId = nu
...
 
         // Admins can do anything.
-        if (self::hasRole(Auth::user(), 'Administrator')) {
+        if (self::hasRole($user, 'Administrator')) {
             return true;
         }
 
+        if (self::hasRole($user, 'NetworkCoordinator')) {
+            $group = Party::find($partyId)->theGroup;
+            foreach ($group->networks as $network) {
```
ngm, line 233: For reference there's a method $user->isCoordinatorForGroup that does the same check (although differently) - I think your method here is a bit more readable though, and possibly more performant, might be worth updating isCoordinatorForGroup to match.

PR #633, opened 2023-03-31
RES-1855 Improve command to sync users for network to Discourse group

edwh:
The old script only sync'd users with `repair_network` set to the network, but that wouldn't work when a network was created after groups/users already existed.

This syncs any users who follow groups in the network.  But it doesn't reset the `repair_network` field - please consider.

### DIFF
--- a/app/Console/Commands/SyncNetworkUsersToDiscourseGroup.php
+++ b/app/Console/Commands/SyncNetworkUsersToDiscourseGroup.php
@@ -56,22 +56,22 @@ class SyncNetworkUsersToDiscourseGroup extends Command
         $network = Network::where('shortname', $networkName)->first();
         $discourseGroupName = $network->discourse_group;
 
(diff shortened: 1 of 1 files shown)


ngm (approved):
Rather than syncing users on their 'repair_network' field, syncs based on group membership to a group that is in the network.

Better, as the old syncing only worked when users were either bulk created for a new network, or joined Restarters through a particular network subdomain.  Which, for the case of existing users from the Cambridgeshire network, neither of which would be true.

This new way will be more robust in general.

ngm:
> This syncs any users who follow groups in the network. But it doesn't reset the repair_network field - please consider.

That's fine.  I think nowadays, we should either:

- only consider repair_network to represent the network that they joined Restarters via
- get rid of it

I think the former is potentially useful, but perhaps we could rename the field to make it clear what it does now.  I'll log it in for more discussion: https://therestartproject.atlassian.net/browse/RESDISCO-163

PR #635, opened 2023-03-31
RES-1880 Ordering of events on dashboard

edwh:
The upcoming events on the dashboard are appearing in an odd order.  This is because we have a default ORDER BY introduced by `scopeUndeleted`, but when that is then used indirectly via other scopes which do a `union`, the order doesn't get applied to the overall query.

Eloquent is in a bit of a guddle over this. There is a `reorder() ` method which you can use to strip out earlier ORDER BY clauses in order to replace it with your own.  But this method doesn't seem to play nice with union queries, so it doesn't fully work for the case we want.

So I've changed one of the scopes to use `reorder` earlier enough so that it works, and then impose the order again in a parent scope.

There are times when it feels like writing raw queries is a lot easier.

### DIFF
--- a/app/Http/Controllers/PartyController.php
+++ b/app/Http/Controllers/PartyController.php
@@ -115,7 +115,7 @@ class PartyController extends Controller
             $group = Group::find($group_id);
         } else {
             // This is a logged-in user's events page.  We want all relevant events.
-            foreach (Party::forUser()->get() as $event) {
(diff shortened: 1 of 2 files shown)


ngm (approved):
Oof, indeed, there comes a point when the ORM always gets in knots.

Workaround works though!

PR #636, opened 2023-04-05
RES-1795 Add API calls for event create/edit

edwh:
Just add the API calls and change the tests for now - the web code still uses the old controller methods until we Vue-ify it.


app/Http/Controllers/API/EventController.php
```diff
@@ -319,29 +354,416 @@ private function getUser() {
...
+
+        $group = Group::findOrFail($groupid);
+
+        // Check whether the event should be auto-approved, if all of the networks it belongs to
+        // allow it.
+        $autoapprove = $group->auto_approve;
+
+        if (!Fixometer::userCanCreateEvents($user))
+        {
+            // REVIEW: This doesn't check that they are a host of this particular group.  That is existing logic.
```
ngm, line 482: Yes, it should here, I think.  Otherwise in theory it could be abused to create an event for a group you don't have rights to.

That said, I wouldn't **necessarily** change it in within userCanCreateEvents, as it's possible that that is a separate check to see whether a user can create events in general, not intended specifically for a given group.  I'm not sure, but we should check.

But ultimately I'd prefer to have a permission check here that is clearly named e.g. `userCanCreateEventsForGroup`

app/Http/Controllers/API/EventController.php
ngm, line 1: Most of the changes to this file seems to be code style / indentation changes, which makes it hard to tell if I've reviewed everything I need to.  I don't think the changes are in line with whatever PSR we follow, either, so probably Laravel Shift will change them back next time we run it.

app/Http/Controllers/API/EventController.php
```diff
@@ -319,29 +354,416 @@ private function getUser() {
...
+            'location' => $location,
+            'latitude' => $latitude,
+            'longitude' => $longitude,
+            'timezone' => $timezone
+        ];
+
+        $party = Party::findOrFail($idEvents);
+        $party->update($update);
+
+        // REVIEW: In the old code we wouldn't generate EditEvent in the approval case.  But I think that was a bug.
```
ngm, line 662: I'm not sure.  Because $party->approve() triggers ApproveEvent, which creates a post on WordPress.  So if we're then triggering EditEvent immediately afterwards, we'd be updating the just created event.  Which might be harmless enough, but I guess there's also a possibility of a race condition if they're queued (which I can't remember if they are)

app/Http/Controllers/API/EventController.php
```diff
@@ -319,29 +354,416 @@ private function getUser() {
...
+        if ($request->has('moderate') && $request->input('moderate') == 'approve') {
+            $party->approve();
+        }
+
+        event(new EditEvent($party, $update));
+
+        if ($request->has('users')) {
+            // REVIEW Existing logic.  But is this needed?  If so we should put it in the OpenAPI spec.  But I
```
ngm, line 670: I think you're right, I certainly have no recollection of us ever passing a list of users.  As far as I know we add/remove users to/from an event one by one when people perform those actions - right?  In which case, this isn't needed.

ngm (approved):

app/Http/Controllers/API/EventController.php
edwh: @ngm My IDE, probably.  Tell me what PSR we are trying to follow and I'll configure it.

app/Http/Controllers/API/EventController.php
edwh: I think this is part of the more general point that Restarters permissions are not well-defined and don't use idiomatic Laravel methods.  So I will change this to a TODO.

app/Http/Controllers/API/EventController.php
edwh: That's what I think.  Have retired this method.

app/Http/Controllers/API/EventController.php
edwh: They are queued, though I would expect that won't cause a race because we have a single queue worker.  There is a window where we will approve on Wordpress and then edit, but I don't think we need to worry too much about that because the Wordpress site should (and will) be using the API to access the data.

Conceptually, I think the point of events is to indicate what has happened, and it is then up to the handler to decide what to do with them.  We shouldn't suppress generation of an event when the thing has happened as a way of stopping the handler doing something, otherwise we lose the point of the interface.

app/Http/Controllers/API/EventController.php
edwh: @ngm PhpStorm has a "Laravel" coding convention, so I've applied that and reformatted.

PR #637, opened 2023-04-11
RES-1863 Add draggable map for group create

edwh:
1. Add a draggable map component.  This will be shared later with events (which isn't in Vue yet).
2. Use this component from group create/edit.
3. Allow the API call to take a lat/lng from the client if passed; otherwise we fall back to the geocoded location.


edwh:
Superseded by #895, which re-implements this on the current codebase (the static location map and Places autocomplete having landed in the meantime, only the draggable fine-tuning and the API lat/lng override remained). The Mapbox geocoder switch that was bundled in here belongs to RES-1858 / #644.

PR #638, opened 2023-04-11
RES-1881 MySQL 8

edwh:
Changes to CircleCI config to run on MySQL 8.

Fix a couple of test bugs which are only exposed in MySQL 8 because it pays attention to timezones in date literals, which earlier versions didn't.


edwh:
We will be upgrading, but at the same time as moving to different hosting, so we're not merging this any time soon.

PR #640, opened 2023-04-19
RES-1878 add user defined fields for group/events

edwh:
PR based on Mapbox branch.

This PR:

1. Adds a `network_data` field to group/event.  This can be set on API calls and is returned in both v1 and v2 APIs.  Add some tests.
2. Adds a reverse geocode method and an artisan command which uses it to set the network data for all groups in the network.  This uses the admin places - looking at the returned data that seems to be the best option.
3. Updates the lat/lng for groups while we're at it.

Some reverse geocodes will fail.


ngm:
I amended the PR to only use Mapbox for the setting of place in network_data for now, keeping the other Mapbox related changes in #634.

I also changed the data we use for the place from Mapbox to more closely relate to the town names that the Repair Cafes are using, and to use the network's default locale for the language.

PR #646, opened 2023-05-05
RES-1891 Network Coordinator can't create events on other groups they created

edwh:
The code in `User::groupsInChargeOf` had bad logic.

### DIFF
--- a/tests/Feature/Events/CreateEventTest.php
+++ b/tests/Feature/Events/CreateEventTest.php
@@ -711,4 +711,30 @@ class CreateEventTest extends TestCase
         # Should have queued ApproveEvent.
         self::assertEquals(0, Queue::size('database'));
     }
+
+    /** @test */
+    public function network_coordinator_other_group() {
+        $network = Network::factory()->create();
+
+        // Create a group in the network.
+        $groupInNetwork = Group::factory()->create();
+        $network->addGroup($groupInNetwork);
+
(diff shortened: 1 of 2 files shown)


tests/Feature/Events/CreateEventTest.php
```diff
@@ -711,4 +711,30 @@ public function notifications_are_queued_as_expected()
...
+        $network->addGroup($groupInNetwork);
+
+        $coordinator = User::factory()->networkCoordinator()->create();
+        $network->addCoordinator($coordinator);
+
+        $this->actingAs($coordinator);
+
+        // Create a group not in the network.
+        $idgroup = $this->createGroup();
+        $groupNotInNetwork = Group::findOrFail($idgroup);
```
ngm, line 730: The code above in User.php looks correct, but this test doesn't look quite right to me - shouldn't the user be made a host of $groupNotInNetwork?

So in theory this test should currently be failing, because this user shouldn't be seeing $groupNotInNetwork as it stands, given it's not in their network and they're not a host of it?

ngm: Oh, does $this->createGroup() create the group on behalf of the current user?

(I looked, and yes, yes it does.  So all good!)

ngm (approved):
Looks good.  Removes the assumption that a network coordinator won't be a host of a group outside of their network.

PR #649, opened 2023-05-09
RES-1888 Category_creation field shouldn't be changed when editing a device

edwh:
Bad code in `ajaxEdit`.  Add a test, and make it work.


ngm (approved):
Looks good, plus good to have the test added, thanks :+1:

PR #650, opened 2023-05-09
RES-1883 Remove useful URLs feature

edwh:
This feature was very rarely used, and has been retired.  This PR removes the code and adds a migration to drop the table.


ngm (changes requested):
Not an insignificant amount of code, for such an unused feature!

I notice that the migration is empty.  I presume we just need to be dropping the device_urls table?

edwh:
> Not an insignificant amount of code, for such an unused feature!
> 
> I notice that the migration is empty. I presume we just need to be dropping the device_urls table?

Added

ngm:
@edwh Did you mean to close this PR?

edwh:
@ngm No, I didn't.  Re-opened so you can review the migration.

PR #651, opened 2023-05-11
RES-1892 Add ability to edit network data for groups.

edwh:
We have a `network_data` field for groups.  This is JSON-encoded.  This PR exposes editing of that for network coordinators on the groups page.

A separate text input is created with a suitable label for each entry in the JSON object.  

There's a button to allow you to add a new (empty) entry, which effectively allows you to extend the schema.  You can't delete them, though, as that is a bit dangerous.

I've done this for groups only, though we may choose to extend it to events too.

### DIFF
--- a/app/Http/Resources/Group.php
+++ b/app/Http/Resources/Group.php
@@ -279,7 +279,7 @@ class Group extends JsonResource
             'tags' => new TagCollection($this->group_tags),
             'timezone' => $this->timezone,
             'approved' => $this->approved ? true : false,
-            'network_data' => $this->network_data,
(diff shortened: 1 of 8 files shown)


ngm (approved):
Great stuff, all looks good to me.

@edwh  Out of interest, why do we need the checks for whether the network_data field is a string or not?

edwh:
> @edwh Out of interest, why do we need the checks for whether the network_data field is a string or not?

I wondered if you'd ask.  I think it was to do with how Laravel handles JSON fields on different database versions.  That should be transparent to the user whether or not it is stored in the DB as a JSON or TEXT field.  But I saw a case in testing when it wasn't, so I put that in and didn't think much further.

PR #654, opened 2023-05-17
RES-1891 NetworkCoordinator sees create event option but can't actually create

edwh:
More bad login in permissions code.

### DIFF
--- a/app/Helpers/Fixometer.php
+++ b/app/Helpers/Fixometer.php
@@ -114,7 +114,7 @@ class Fixometer
             }
         }
 
-        if (self::hasRole($user, 'Host') && self::userIsHostOfGroup($group->idgroups, $userId)) {
+        if (self::userIsHostOfGroup($group->idgroups, $userId)) {
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


app/Helpers/Fixometer.php
```diff
@@ -140,21 +141,15 @@ public static function userHasEditPartyPermission($partyId, $userId = null)
...
         }
 
-        if (self::hasRole($user, 'Host')) {
-            $group_id_of_event = Party::where('idevents', $partyId)->value('group');
-            if (self::userIsHostOfGroup($group_id_of_event, $userId)) {
-                return true;
-            } elseif (empty(DB::table('events_users')->where('event', $partyId)->where('user', $user->id)->first())) {
-                return false;
-            }
+        if (self::userIsHostOfGroup($group->idgroups, $userId)) {
```
ngm, line 151: That removed logic looks a bit odd - it looks like it was once attempting to check whether the user has any kind of association to the event?  Makes me wonder if elsewhere is using it to check whether someone can edit the items at an event.

But then essentially it looks like it was a no-op because it was returning false.

So as I'm guessing that the tests are all still passing, then this change looks good :+1:

ngm: Ah right, so it looks like it was trying to ensure that a 'Host' who was not a host of that group not able to edit the party.  Which is exactly what userIsHostOfGroup does.  So all good!

ngm (approved):

PR #655, opened 2023-05-22
RES-1894 Allow country names to be translated.

edwh:
And translate Belgium in fr-BE as an example.

1. Move list of countries into a new translation file.  
2. The existing `getAllCountries` method now gets this using the `Lang` interface and takes a `locale` parameter for reasons which will become clear shortly.
3. We continue to store countries in the DB as names.  
4. Therefore when we are returning a country name we need to translate it using the current locale, which is where the `locale` parameter gets used.  Introduce a `translateCountry` method and use it.
5. When we are creating a group, the country comes from the geocoder and is therefore in English.  So we don't need to reverse translate.
6. The API should (I claim) continue to return the values in English by default.  But to allow the client to show the translated values, introduce a `locale` parameter which can be used on all API calls, and middleware to access this and set the app locale.
7. Add this parameter on the client.

I'm not completely happy with this function.  

1. We should consider changing the countries stored in the DB to country codes.  The `array_flip` stuff is a smell.
2. I can't decide whether the `locale` parameter smells.


app/Http/Middleware/SetLocale.php
ngm, line 1: I note that there's already a (badly named) middleware called LanguageSwitcher.php.  It looks like there's some duplication of functionality and that has more functionality.  Should that one be used / renamed / should they be merged?

ngm:
I agree with storing the countries in the DB by their country code.  If I had to pick between 2 char and 3 char of ISO 3166  I would go with 3 char but I would imagine if we did it, we would use whatever Mapbox gives us back.

I would supplement the country name column with an additional country code column, rather than replace it, so dependent code / reports etc aren't affected.

As discussed having a locale query string parameter that can be added to the API makes sense to me, but as noted in the review comments, we have something that (looks like it? does it work for API?) does that already in LanguageSwitcher.php so we should stick with that or merge them so we're not doing something with that parameter twice.

ngm (changes requested):
Looks good in general, though see Jira ticket - it wasn't translated everything for me.

I agree would be add country codes to the DB.

Also could you check LanguageSwitcher and see if anything it is doing is relevant to SetLocale middleware?

app/Http/Middleware/SetLocale.php
edwh: I'm not keen to add a separate column in parallel to the existing country column. Single source of truth, surely?  It's basically just increasing tech debt to do that, I think.  It would be better to bite the bullet, but perhaps in a follow-on story.  Let's discuss.

I hadn't noticed that the `LanguageSwitcher` middleware includes the same kind of function.  But yes, this is only used for `web` routes, not for `api` routes.  The new `SetLocale` middleware is for the API.  I'll rename that to `APISetLocale` to make that a little clearer - ok?

In due course `LanguageSwitcher` ought to die when the client is using the API throughout.

ngm: > I'm not keen to add a separate column in parallel to the existing country column. Single source of truth, surely? It's basically just increasing tech debt to do that, I think. It would be better to bite the bullet, but perhaps in a follow-on story. Let's discuss.

Fair point - but in which case agree it should be a separate ticket, as it'll need more time to make the change.

>  I'll rename that to APISetLocale to make that a little clearer - ok?

Sounds good, thanks.  Just for clarity in my head, I'm also seeing a distinction that APISetLocale is also only setting the locale for the current request, correct?  Whereas when LanguageSwitcher sets the locale, it's also setting it for the user more permanently.

edwh: > Sounds good, thanks. Just for clarity in my head, I'm also seeing a distinction that APISetLocale is also only setting the locale for the current request, correct? Whereas when LanguageSwitcher sets the locale, it's also setting it for the user more permanently.

Yes, that's as intended.

PR #658, opened 2023-06-01
RES-1896 Change "New Zealand" to "Aotearoa New Zealand"

edwh:


ngm (changes requested):
I assume this will need to be changed based on what is in RES-1894?

edwh:
> I assume this will need to be changed based on what is in RES-1894?

I thought we were just changing the NZ country name throughout (without prejudice to other political disputes) so that we didn't have to worry about how to get the relevant locale set for a network which doesn't have a vanity domain?

So I think RES-1894  is separate,

ngm:
Yes to all the above, but in RES-1984 I saw that the countries list had moved out of Fixometer and into Lang files.

edwh:
> Yes to all the above, but in RES-1984 I saw that the countries list had moved out of Fixometer and into Lang files.

Ah, yes.  You're right, this will be a different fix once that goes in.  Merge time.

ngm (approved):

PR #661, opened 2023-06-05
RES-1906 Demote network coordinator

edwh:
When we demote someone from a role of network coordinator, we should remove them from any networks they are currently coordinating.

Some test fiddling to make it easier to run them in isolation.

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -380,9 +380,9 @@ class UserController extends Controller
             'role' => $request->input('user_role'),
         ]);
 
-        // If we are demoting from NetworkCoordinator to host, remove them from the list of coordinators for
+        // If we are demoting from NetworkCoordinator, remove them from the list of coordinators for
         // any networks they are currently coordinating.
-        if ($oldRole == Role::NETWORK_COORDINATOR && $user->role == Role::HOST) {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -373,11 +373,19 @@ public function postAdminEdit(Request $request)
...
+
         // Set role for User
         $user->update([
-        'role' => $request->input('user_role'),
+            'role' => $request->input('user_role'),
         ]);
 
+        // If we are demoting from NetworkCoordinator, remove them from the list of coordinators for
+        // any networks they are currently coordinating.
+        if ($oldRole == Role::NETWORK_COORDINATOR && ($user->role == Role::HOST) || $user->role == Role::RESTARTER) {
```
ngm, line 385: Is the operator precedence OK here?  Might be misreading it, but looks like anyone who is a Restarter will get detached from the network.

ngm (changes requested):
Just double-checking the logic.

app/Http/Controllers/UserController.php
edwh: It's wrong.  Thank you.  Fixed.

ngm (approved):

PR #662, opened 2023-06-05
RES-1902 Change placeholder for brand/model for when it's not known

edwh:
Change the placeholders.  Add a list of "unknown" synonyms and wipe the brand/category value if it matches.


ngm (approved):
LGTM :+1:

ngm (approved):

PR #665, opened 2023-06-21
RES-1908 country codes

edwh:
- Add the `country_code` field to users and groups via a migration.
- Change the code to use the `country_code` field.  I initially did this with the migration deleting the `country` field in the `groups` table to get the tests running without it, then removed that delete.
- Add a scheduled hourly command to sync the `country code` field to `country`


database/migrations/2023_06_12_153317_country_codes.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
ngm, line 54: Missing a reverse of the migration?

ngm (approved):
All looks good, just missing a reverse of the migration I think.

I'll have a bit of a smoke test through locally.

PR #668, opened 2023-07-10
WIP RES-1910 event chat counts

edwh:
Based on RES-1855 so the diffs are confusing until that is merged.

- Move the maintenance of the volunteer count and the Discourse thread into an observer, which is triggered by changes to EventsUsers.  This means we don't have to rely on the calling code to remember to do it.
- Background the actual processing this via events and listeners in the usual way
- When the host of a group is added to an event, they should become a host of the event.
- Allow fetching of an event over the API if it is not yet approved.  No reason not to do this, and it helps with testing.

TODO Doesn't yet remove from the Discourse thread.


app/Http/Controllers/API/EventController.php
```diff
@@ -161,9 +161,18 @@ public function addVolunteer(Request $request, $idevents)
...
+
         if ($request->has('user') && $request->input('user') !== 'not-registered') {
+            // User is null, this volunteer is either anonymous or no user exists.
             $user = $request->input('user');
+
+            if ($user) {
+                $u = User::find($user);
+
+                // A host of a group who is added to an event becomes a host of the event.
+                $eventRole = $u && $u->role == Role::RESTARTER ? Role::RESTARTER : Role::HOST;
```
ngm, line 174: This would make network coordinators and admins hosts of the event, too, right?

I can think of cases where that's not going to match reality (e.g. me, going to an event).  But, probably generally fine.

ngm: For the case where a host of one group attends an event by another group, this would automatically make them host of that event?  I don't think that should be the case.  It should only turn hosts of the event's group into hosts of the event, no?

ngm (commented):
Couple of questions about the logic of who is set as a host of an event when joining the event.

app/Http/Controllers/API/EventController.php
edwh: > This would make network coordinators and admins hosts of the event, too, right?
> 
> I can think of cases where that's not going to match reality (e.g. me, going to an event). But, probably generally fine.

Would it?  They'd have `Role::NETWORK_COORDINATOR` or `Role::ADMIN`.

edwh: The second point would also fix the first.  This logic isn't well-tested in any case, so I've added a test for it to check that people are/aren't added as hosts appropriately.

app/Http/Controllers/API/EventController.php
```diff
@@ -161,9 +161,18 @@ public function addVolunteer(Request $request, $idevents)
...
+
         if ($request->has('user') && $request->input('user') !== 'not-registered') {
+            // User is null, this volunteer is either anonymous or no user exists.
             $user = $request->input('user');
+
+            if ($user) {
+                $u = User::find($user);
+
+                // A host of the group who is added to an event becomes a host of the event.
+                $eventRole = $u && $u->role == Role::HOST && Fixometer::userIsHostOfGroup($party->group, $user) ? Role::HOST : Role::RESTARTER;
```
ngm, line 174: I think this should just be

`$eventRole = $u && Fixometer::userIsHostOfGroup($party->group, $user) ? Role::HOST : Role::RESTARTER;`

as there could be admins or network coordinators who are host of the group, and should get set as host of event, even if their overall role might not be Role::HOST.

ngm (changes requested):
I think the logic for setting event role needs changing, see comment.

ngm (approved):

PR #669, opened 2023-07-10
RES-1913 Group table filter bugs.

edwh:
Reinstate a bit of JS which was over-zealously tidied away.

Fix location filter.

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -734,6 +734,34 @@ function initAutocomplete() {
       }
     });
 
+    function removeUser() {
+
+      var id = jQuery(this).data('remove-volunteer');
+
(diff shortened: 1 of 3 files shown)


ngm (approved):
@edwh How come the removal of the removeUser method caused the errors in the group table filters?  Regardless, given it works, all good with the change :+1:

edwh:
> @edwh How come the removal of the removeUser method caused the errors in the group table filters? Regardless, given it works, all good with the change 👍

If you get a JS error early on, then it blocks the load of later JS, and therefore that JS never fires.  It isn't always obvious which kinds of JS errors cause that but any console error is worth investigating.

PR #672, opened 2023-08-07
RES-1926 Event handler is looking for data which doesn't get passed

edwh:
`EditWordpressPostForEvent` is looking in the data passed to the (Laravel) event for the id of the group.  That isn't always present, and doesn't need to be because we have the (Restarters) event in hand and can pull the group from there.


ngm (approved):
As noted in Jira, I think this may mean that event group changes are no longer propagated, but that that's OK.

PR #696, opened 2023-10-16
RES-1949 Add API for group volunteers

edwh:
1. Add API calls to list, promote and remove group volunteers.
2. Use these from within the Vue code rather than passing the volunteers from the blade.  
3. Modify the `volunteers` store, including having an eye to the future when we will add similar code for events.
4. Remove the "nearby volunteers" function rather than updating it.


app/Http/Controllers/API/GroupController.php
```diff
@@ -413,32 +414,170 @@ public static function getEventsForGroupv2(Request $request, $idgroups) {
...
-        $user = auth('api')->user();
+    public function deleteVolunteerForGroupv2(Request $request, $id, $iduser)
+    {
+        $user = $this->getUser();
 
-        if (!$user || !Fixometer::userHasEditGroupPermission($idgroups, $user->id)) {
-            // We require host permissions to view the list of volunteers.  At the moment this call is only used when
-            // adding volunteers, and this check means we don't have to worry about exposing sensitive data.
-            abort(403);
+        list($name, $area, $postcode, $location, $phone, $website, $description, $timezone, $latitude, $longitude, $country, $network_data, $email) = $this->validateGroupParams(
```
ngm, line 501: Why do we need this line out of interest?

app/Http/Controllers/API/GroupController.php
```diff
@@ -413,32 +414,170 @@ public static function getEventsForGroupv2(Request $request, $idgroups) {
...
+     *      ),
+     *     )
+     */
+
+    public function patchVolunteerForGroupv2(Request $request, $id, $iduser)
+    {
+        $user = $this->getUser();
+        $host = $request->get('host', false);
+
+        list($name, $area, $postcode, $location, $phone, $website, $description, $timezone, $latitude, $longitude, $country, $network_data, $email) = $this->validateGroupParams(
```
ngm, line 562: Same question as above.

ngm (approved):

app/Http/Controllers/API/GroupController.php
edwh: We don't.  Thanks.

app/Http/Controllers/API/GroupController.php
edwh: Same answer as above.

PR #705, opened 2023-11-27
RES-1954 Add banner for Aviva crowdfunder

edwh:
As subject.


resources/js/components/DashboardBanner.vue
ngm, line 1: Banner looks good - but think it needs the id of the banner changing on line 33?  Currently set to biggive2023.

ngm (changes requested):

ngm (approved):

PR #737, opened 2024-08-05
RES-1976 Tidy up old Discourse log files

edwh:
We're using Monolog for Discourse API logging, which doesn't have built-in compression and deletion - see https://github.com/Seldaek/monolog/issues/1202

This PR is a way of deleting old log files.  It feels a bit clunky - there might be a better way to do it - but it works.


ngm:
I'm fine with this, but for reference I believe there is a 'days' parameter on the 'daily' log driver, which you can use to specify many days of logs to keep.

https://betterstack.com/community/guides/logging/how-to-start-logging-with-laravel/#understanding-channel-drivers

ngm (approved):

edwh:
I think you're right, and it would just be tech debt to do it the way I've done it.  So I'll rework.

ngm (approved):

PR #743, opened 2024-09-09
RES-1992 add ability to archive inactive groups

edwh:
- [x] Add a field in a group which records the `archived_at` date
- [x] Expose this value in the API.
- [x] Allow setting it by Network Coordinator or Admins over the API
- [x] Exclude archived groups from lists returned by the API unless an `archived` parameter is passed.
- [x] Add option in Group Actions dropdown to archive a group, only visible to admins and network coordinators.
- [x] See archived groups in listings, with a Bootstrap badge to indicate status.  Applies to dashboard, groups list, group page.  NB follow-on work will hide these groups for users who are not Network Coordinators or Admins.
- [x] Suppress notifications for archived groups.
- [ ] Migrate group tag/titles to `archived_at` flag.


app/Http/Controllers/API/GroupController.php
```diff
@@ -252,14 +261,25 @@ public static function getGroupList()
      */
 
     public static function listNamesv2(Request $request) {
+        $request->validate([
+            'archived' => ['string', 'in:true,false'],
```
ngm, line 265: I'd probably call this parameter includeArchived, as when it's called 'archived', passing archived=true could make one think you would only be receiving archived.  Elsewhere we have 'includeDetails' so perhaps better for consistency, too.  That said, I can live with it if simpler to keep it.

database/migrations/2024_09_16_093650_inactive_groups.php
```diff
@@ -0,0 +1,45 @@
...
+        $groups = \App\Group::join('grouptags_groups', 'groups.idgroups', '=', 'grouptags_groups.group')
+            ->join('group_tags', 'grouptags_groups.group_tag', '=', 'group_tags.id')
+            ->where('group_tags.id', \App\GroupTags::INACTIVE)
+            ->get();
+
+        foreach ($groups as $group) {
+            $group->archived_at = $group->updated_at;
+
+            // Remove [INACTIVE] from the group name - this is now indicated via archived_at.
+            $group->name = str_replace('[INACTIVE]', '', $group->name);
```
ngm, line 25: This should also remove the space, i.e. '[INACTIVE] '.

ngm (changes requested):
The migration needs amending, and I can take or leave my other comment.  

See also a couple of things mentioned in Jira ticket.

database/migrations/2024_09_16_093650_inactive_groups.php
edwh: Agree.

PR #773, opened 2024-12-18
RES-2001 anonymise Discourse users when Restarters user deleted

edwh:
This adds an event listener which will use Discourse's API to anonymise Discourse users as their account is soft-deleted on Restarters.

There's also an `artisan` command to anonymise an individual user.  We'd need to manually run this for the live soft-deleted users - easy enough as a one off.  I've not written a migration script.

### DIFF
--- a/app/Listeners/DiscourseUserEventSubscriber.php
+++ b/app/Listeners/DiscourseUserEventSubscriber.php
@@ -117,6 +117,22 @@ class DiscourseUserEventSubscriber extends BaseEvent
         }
     }
 
+    public function onUserDeleted(UserDeleted $event)
+    {
+        if (config('restarters.features.discourse_integration') === true)
+        {
+            $user = $event->user;
+
+            try
+            {
+                $this->discourseService->anonymise($user);
+            } catch (\Exception $ex)
+            {
+                Log::error('Could not anonymise ' . $user->id . ' on Discourse: ' . $ex->getMessage());
+            }
+        }
+    }
+
     /**
      * Register the listeners for the subscriber.
      *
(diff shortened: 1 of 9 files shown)


app/Listeners/DiscourseUserEventSubscriber.php
```diff
@@ -117,6 +118,24 @@ public function onUserRegistered(UserRegistered $event)
...
+    public function onUserDeleted(UserDeleted $event)
+    {
+        if (config('restarters.features.discourse_integration') === true)
+        {
+            $user = $event->user;
+
+            try
+            {
+                $this->discourseService->anonymise($user);
+                $user->username = null;
```
ngm, line 130: I don't think we should be nulling the username here.  It's anonymised in app/User.php:368.

ngm (changes requested):

app/Listeners/DiscourseUserEventSubscriber.php
edwh: I'd missed that, thanks.  But I don't think it's right for us to put in a username that won't match what it is on Discourse, and it's not for us to presume what that will be.  So I think it's better to remove the setting of it in `User.php`.  What do you think about that?

ngm: I don't think it's the end of the world if the usernames get out of sync, as presumably the id in Laravel and the external id in Discourse will still match.  But I agree, if straightforward, would make sense to have the usernames consistent between the two.

What happens to the username in Discourse though when you anonmise?  Is it nulled or set to something anonymous?

If it's nulled, then perhaps we should also null in Laravel.  However I would not be confident that nulling the username in Laravel won't break some page somewhere that isn't checking for nulls.  I imagine we don't have test coverage for it.  The users are soft deleted so perhaps its not an issue.  

If it's anonymised in Discourse, then perhaps we could pull that back and set the Laravel username to the same.  Or vice versa.

edwh: It's anonymised in Discourse, and it handily returns the new name.  So I've made it store that instead.

ngm (approved):

PR #775, opened 2025-01-07
RES-2008 Add reference field for devices

edwh:
Wait till nuxt deployed then reimplement.

Add this to the device API, device add/edit, and also show (subtly) in the device summary.

### DIFF
--- a/app/Device.php
+++ b/app/Device.php
@@ -67,7 +67,7 @@ class Device extends Model implements Auditable
      *
      * @var array
      */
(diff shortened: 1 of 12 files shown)


edwh:
Reimplemented on the Nuxt 4 branch in #907, so this doesn't get lost when the SPA lands.

The API side there is your change as-is (migration, `$fillable`, validation, resource, OpenAPI schema, both create and update). The client side had to be rebuilt against the SPA's components, where the reference sits with brand/model in the device form and shows small and muted above the item type in the device tables — same idea as your `EventDeviceSummary.vue` change.

Leaving this open, since it still applies to the current stack. Thanks for the original — the criteria and the wording of the tooltip carried straight over.

PR #778, opened 2025-01-15
RES-2012 Speed up /api/v2/networks/N/groups call

edwh:
This call is slow due to collection of statistics.  This PR defaults statistics off unless `includeStats` is set true on the request.


edwh:
Implemented separately.

PR #783, opened 2025-02-20
RES-2015 Host can't create event

edwh:
The function to change the role of a member to Host was lost in recent API changes.  This restores it and adds a test.


ngm (commented):
:+1:

PR #787, opened 2025-03-06
RES-2020 retire Workbench

edwh:
There are no plans for any more data collection quests.  This PR removes the navbar, page, and related code.  The tables are renamed in the DB for archiving purposes.


database/migrations/2025_03_06_110500_retire_workbench.php
```diff
@@ -0,0 +1,78 @@
...
+        'microtask_surveys',
+    ];
+
+    public function up()
+    {
+        // For each table, rename to 'archived__' + table name.
+        // Using two underscores triggers phpMyAdmin table grouping.
+        foreach ($this->tables as $table) {
+            try {
+                Schema::rename($table, 'archived__' . $table);
```
ngm, line 54: Could we perhaps have them as something like 'z_archived__'?

Just a minor point, as having all of these tables at the beginning of the tables list in phpmyadmin or reporting software will be a slight inconvenience.  That said, I can hide them in e.g. Metabase, so no real problem there.  

I don't mind too much.  What do you think?

lang/en/devices.php
```diff
@@ -29,7 +29,7 @@
   'global_impact' => 'Our Global Impact',
   'group_prevented' => 'prevented <a href="/party/view/:idevents">:amount kg</a> of waste!',
   'huge_impact' => 'Repairers around the world are having a huge impact!',
-  'impact_read_more' => '<a target="_blank" href="https://therestartproject.org/faq">Read more</a> about how we assess our impact on the environment. You can make our impact go even further. Help us understand what else this data tells us and what changes we should campaign for over at our <a href="/workbench">Data Workbench</a>.',
+  'impact_read_more' => '<a target="_blank" href="https://therestartproject.org/faq">Read more</a> about how we assess our impact on the environment.',
```
ngm, line 32: This is fine, though noting there's a slight additional change to this in https://therestartproject.atlassian.net/browse/RES-2024 .

resources/js/components/DashboardPage.vue
```diff
@@ -12,12 +12,6 @@
       </div>
       <DashboardAddData class="adddata justify-self-end" />
       <DashboardRightSidebar class="sidebar" />
-      <DiscourseDiscussion
```
ngm, line 15: This component shouldn't be removed from the main dashboard.  (In this ticket, at least...).

resources/js/components/DiscourseDiscussion.vue
ngm, line 1: This component should remain for now (used on the main dashboard).

ngm (changes requested):

database/migrations/2025_03_06_110500_retire_workbench.php
edwh: phpMyAdmin groups them, though, so it's just one line.  I kind of like being obliged to notice we have crud in the DB.  But ask me again and I'll change it.

resources/js/components/DashboardPage.vue
edwh: Agree, not sure why I zapped that.

database/migrations/2025_03_06_110500_retire_workbench.php
ngm: Ah yeah, good point about PMA.  And also agree, having it visible pushes one to remove it.  So, fine as is!

PR #795, opened 2025-04-03
Feature: Improve Development Setup

ardelato:
## Description

This pull request introduces two major changes to enhance the local development setup for the Restarters project.

The changes include:
- The addition of a new `Taskfile.yml` for task management
- Updating the `docker-compose.yml` file to support different development profiles

In addition, I've added documentation surrounding these changes.

### CR Notes:

Commit https://github.com/TheRestartProject/restarters.net/commit/703d8b6317086b282f18acb334c52857235ced18 goes over the reason for choosing [Task](https://taskfile.dev/) as opposed to a `bash` script or `Makefile`. 

By utilizing [Docker Compose Profiles](https://docs.docker.com/compose/how-tos/profiles/) we can customize which services we would spin up locally. That way we do not utilize more resources than needed for local development.

As for the documentation, it goes over the prerequisites and setup for local development, as well as examples of how to run frequent actions. That said, I added the document to `/docs` but I am not sure if that is the correct location for it.


docs/local-development.md
```diff
@@ -0,0 +1,226 @@
...
+
+### 3. Starting the Development Environment
+
+The development environment includes several optional services that can be enabled through profiles:
+
+- **Core**: Restarters web application and MySQL database
+- **Development**: Adds phpMyAdmin and Mailhog
+- **Discourse**: Adds Discourse forum with its required services (PostgreSQL, Redis, Sidekiq) 
+
+#### Using Task (Recommended)
```
edwh, line 67: Is there any good reason not to make Task mandatory?  Otherwise we have to document/maintain instructions for using Task and not using Task.  Once we've started using it, we'll probably use it for other non-Docker things - so let's just make that decision now.

docs/local-development.md
```diff
@@ -0,0 +1,226 @@
...
+### 2. Environment Configuration
+
+Copy the example environment file:
+
+```bash
+cp .env.example .env
+```
+
+You may need to adjust the following settings in your `.env` file:
+- Set `APP_URL` to `http://localhost:8001` (matching the port in docker-compose.yml)
```
edwh, line 56: Any reason not to change the version in .env.example?  

It's used in `travis.yml`, but we can delete that file - we use CircleCI now.

It's also used in `docker_run.sh`

docs/local-development.md
```diff
@@ -0,0 +1,226 @@
...
+# Using Task
+task docker:logs
+
+# Or using Docker
+docker logs -f restarters
+```
+
+### 5. Accessing the Application
+
+Once setup is complete:
```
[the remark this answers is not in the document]
edwh: When I ran this on Windows, it didn't seem to start the Restarters container correctly - no `docker_run` was running and it wasn't listening on the port.  Digging a bit, the composer install is failing because of PHP version/composer.json mismatches.  

You've not changed the `Dockerfile` so this is probably broken in develop - by me, therefore.  But would you expect this to work yet?  I don't really want to merge this PR until I can start a working test system using it, because that's a good test of this PR.

docs/local-development.md
```diff
@@ -0,0 +1,226 @@
...
+cp .env.example .env
+```
+
+You may need to adjust the following settings in your `.env` file:
+- Set `APP_URL` to `http://localhost:8001` (matching the port in docker-compose.yml)
+- The `APP_KEY` will be auto-generated during setup
+
+### 3. Starting the Development Environment
+
+The development environment includes several optional services that can be enabled through profiles:
```
edwh, line 61: Should we use the names the profiles actually have, e.g. `dev` not `Development`?

edwh (changes requested):
Generally this is looking excellent.  But a few comments and I'm not sure it's ready to merge yet.

docs/local-development.md
ardelato: > Is there any good reason not to make Task mandatory?

The main reason I didn’t push for that right away was to avoid making too many changes at once. I wasn’t sure how receptive folks would be to immediately requiring Task, so I opted to support both paths initially and document how to use it without Task. That way, it’s a smaller transition.

In the end, I did have in mind to fully switch to requiring Task once we had broader use cases beyond Docker.

ardelato: That said, I can update the docs to make `Task` mandatory.

docs/local-development.md
ardelato: >Any reason not to change the version in .env.example?

Are you referring to changing `APP_URL` to the following?
```diff
- APP_URL=http://restarters.test:8000
+ APP_URL=http://localhost:8001
```

If so, then I didn't change it because I lack knowledge.

>It's used in travis.yml, but we can delete that file - we use CircleCI now.
>It's also used in docker_run.sh

As you mentioned, the `.env` file, and thus `APP_URL`, is used in both these files. However, I'm not sure why `APP_URL` was set to that value in the first place as the original commit did not have much information.

https://github.com/TheRestartProject/restarters.net/commit/47afb56355130734cc9bfe976db86e00bde43c3a

I opted to leave it alone since I lacked insight.

docs/local-development.md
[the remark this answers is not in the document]
ardelato: > Digging a bit, the composer install is failing because of PHP version/composer.json mismatches.

Can you try removing this line of code from the `docker-compose.yml` file and see it works now?

https://github.com/iFixit/restarters/blob/23430822afb0462ad510dafecf5c8402d42a0f27/docker-compose.yml#L85

We set an image of `php:7.4` for the Restarters app in the `docker-compose.yml` file. Which is a different version then the `Dockerfile` base image of `cimg/php:8.1.12-node`

That could explain the mismatch errors.

docs/local-development.md
edwh: Change it, please, and delete `travis.yml`.  We'll deal with any issues that creates on our side.

docs/local-development.md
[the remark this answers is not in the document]
edwh: Yes, removing that helps.

I also needed to create a `.dockerignore` file as per https://github.com/docker/for-win/issues/14083#issuecomment-2447288859.  Assuming that is ok by you, can you add it to this PR?

docs/local-development.md
edwh: My concern here is that if you're not familiar with this stuff, we have several similar but different sets of terminology which introduce potential confusion and cognitive load.

I'll roleplay a confused developer who is coming to this cold:
* Core/Development/Discourse ("ok, sounds reasonable")
* up/up-dev/up-discourse/up-all ("hmmm, probably up = core, but how come there are four commands here when I've just read about only three?")
* "basic environment"/"development tools"/"discourse integration"/"all services" ("Ok, the descriptions and capitalisation are a bit different from the three bullet points earlier, but I guess I understand this only I'm not sure if the Discourse flavour includes phpMyAdmin and Mailhog or not, and if it does, what the difference is between up-discourse and up-all").

We should choose one set of terminology and stick to it more ruthlessly.  So if core = Restarters web application + MySQL database, then we should have `task docker:up-core`.

edwh:
Couple of other things from testing.

1. On Windows, I see this log

image

Possibly for this reason, when I changed some of front-end code (e.g. `DashboardPage`) and refreshed, the change never got picked up.  I can't remember if this was a problem with our original docker setup - it's been a while since I used it.  Do code changes work for you?

2. You might not have noticed that to pick up translation changes, you need to:
* Run `php artisan lang:js --no-lib resources/js/translations.js`
* Restart the node watch.  
This is an annoying consequence of sharing the same translations between Laravel blade templates and front-end Vue code.  You may need to add something to Task for that.

FYI: Often during development I run with `APP_DEBUG` set `FALSE` in `.env` because otherwise things are pretty slow - the Laravel overhead which gets all the logs through to the client.

docs/local-development.md
ardelato: Addressed in https://github.com/TheRestartProject/restarters.net/pull/795/commits/90433d011ff932428dad8557a2db5abd028df834

docs/local-development.md
ardelato: Addressed in https://github.com/TheRestartProject/restarters.net/pull/795/commits/a9a619cd6f03243766617be3dc44ba56ce6d00d9

docs/local-development.md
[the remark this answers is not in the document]
ardelato: Addressed in https://github.com/TheRestartProject/restarters.net/pull/795/commits/b234848553bd60a0a4d0fc22e027534372a425d9 and https://github.com/TheRestartProject/restarters.net/pull/795/commits/404ccdb7c714dffb4027f5a11892ec3f3f0cfe04

ardelato:
@edwh I tested it on my Windows machine but I did not see the `git` error you saw. What Docker version do you have and are you using Docker Desktop or just Docker CE?

That said, I have seen this error before and applied the following patch. https://github.com/TheRestartProject/restarters.net/commit/4d1842c453b3f7eef0ae189cd2e17cbb1186f340

This was needed for a Fedora machine we have that is a running an older version of Docker (v24.0.5)

-----


As for the changes not being picked up, I found the following that could explain the underlying cause and have confirmed a solution:

- https://forums.docker.com/t/in-docker-windows-nodemon-does-not-detect-file-changes/135042/2
- https://syntackle.com/blog/the-issue-of-watching-file-changes-in-docker/#the-problem

Basically, Laravel's Mix, aka Webpack, is not receiving the file change events. Therefore, if we instead use [Mix's polling feature](https://laravel-mix.com/docs/6.0/cli#polling), then we can pick up the changes like that.

We already have the poll script in the `package.json` - https://github.com/TheRestartProject/restarters.net/blob/29224072032d743293c5814ca6a2dd74b4c1ec70/package.json#L7 so if that is used instead of `[npm watch](https://github.com/TheRestartProject/restarters.net/blob/29224072032d743293c5814ca6a2dd74b4c1ec70/docker_run.sh#L40)` then the `Vue` changes will be picked up.

I have confirmed this worked on my Windows machine.

docs/local-development.md
```diff
@@ -0,0 +1,157 @@
...
+
+```bash
+cp .env.example .env
+```
+
+You may need to adjust the following settings in your `.env` file:
+- The `APP_KEY` will be auto-generated during setup
+> [!NOTE]
+> This is only required if another service is already running on the port `8001`
+- Set `APP_URL` to `http://localhost:<AVAILABLE_PORT>`
```
edwh, line 51: I think you only need to do this if the default (8001?) is not available.  Better to word it like that.

docs/local-development.md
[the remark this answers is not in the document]
edwh: What about this bit:

Is there any way to make the task not return until everything is ready? Normally I just wait a bit, but that's a bit lame.

docs/local-development.md
```diff
@@ -0,0 +1,157 @@
...
+task docker:up-core
+
+# Start with debug tools (phpMyAdmin, Mailhog)
+task docker:up-debug
+
+# Start with Discourse integration
+task docker:up-discourse
+
+# Start all services
+task docker:up-all
```
edwh, line 75: This is better, thank you.  But it's still not clear what the difference is between `up-all` and `up-discourse`.

edwh:
> @edwh I tested it on my Windows machine but I did not see the `git` error you saw. What Docker version do you have and are you using Docker Desktop or just Docker CE?
> 
> That said, I have seen this error before and applied the following patch. [4d1842c](https://github.com/TheRestartProject/restarters.net/commit/4d1842c453b3f7eef0ae189cd2e17cbb1186f340)
> 
> This was needed for a Fedora machine we have that is a running an older version of Docker (v24.0.5)

I'm running Docker Desktop 4.30.0, which is pretty recent.  I tried added that change to the Dockerfile and I still see the error.  

If you don't see it, then I will need to debug though I'm not sure I can see any consequences of it yet, because...

> I have confirmed this worked on my Windows machine.

Me too.  Can we make this the default, then?

docs/local-development.md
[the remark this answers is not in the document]
ardelato: > Is there any way to make the task not return until everything is ready? Normally I just wait a bit, but that's a bit lame.

There's not much that can be done out of the box. We'd need to incorporate some sort of health check and then incorporate that into the Taskfile.

Is the main goal to get more transparency on the status of the running containers? Would
`task docker:up-core && task docker:logs` suffice?

docs/local-development.md
ardelato: Within the documentation or the task description, or both?

https://github.com/iFixit/restarters/blob/d790c85ce17c3955346faf8ca9f48652a5f8dbd1/Taskfile.yml#L37-L38

I am not really sure what would be the best solution for this.

ardelato:
Did you follow similar steps as outlined in this blog post? https://medium.com/@suyashsingh.stem/increase-docker-performance-on-windows-by-20x-6d2318256b9a

If so, then I can confirm that the Docker build speed and up speed were a lot faster. However, I came across a different issue which has to do with ownership of the mounted volumes and thus get a permission denied error when trying to run `npm run watch&` in the `docker_run.sh` script. This prevented me from validating if changed files were getting picked up.

docs/local-development.md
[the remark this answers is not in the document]
edwh: > Is the main goal to get more transparency on the status of the running containers? Would
> `task docker:up-core && task docker:logs` suffice?

I'm thinking of:
* Someone new to the codebase who wouldn't know which logs meant things were ready, especially once you throw multiple different containers into the mix.
* Use of Docker in CircleCI where we don't want to run tests until the system is up and stable.

It's only a nice to have, though - you can just wait for a bit and see.  So it doesn't need to be as part of this PR.

docs/local-development.md
edwh: Within the documentation.  Think of it this way: what containers does `all` start that `discourse` doesn't?  How would I know that?

edwh:
Coming back to it, I remembered that I needed to do an `su` from within WSL.  Does that help with your permissions issue?

The change to use `watch-poll` still needs to go in, I think.  Otherwise this is looking releasable to me.

ardelato:
@edwh could you share your WSL setup and steps you've taken? I can't seem to properly get the same results you are getting.

Even after using `su` I still get the same permissions errors

```
npm ERR! code 1
npm ERR! path /var/www/node_modules/playwright
npm ERR! command failed
npm ERR! command sh -c -- node install.js
npm ERR! node:internal/process/promises:288
npm ERR!             triggerUncaughtException(err, true /* fromPromise */);
npm ERR!             ^
npm ERR! 
npm ERR! [Error: EACCES: permission denied, mkdir '/root/.cache/ms-playwright'] {
npm ERR!   errno: -13,
npm ERR!   code: 'EACCES',
npm ERR!   syscall: 'mkdir',
npm ERR!   path: '/root/.cache/ms-playwright'
npm ERR! }
npm ERR! 
npm ERR! Node.js v18.12.0

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2025-05-15T13_20_10_292Z-debug-0.log
rebuilt dependencies successfully
Created: resources/js/translations.js

   INFO  Application key set successfully.  


> watch-poll
> mix watch -- --watch-options-poll=1000

sh: 1: mix: Permission denied
```

docs/local-development.md
ardelato:
```diff
 - The `APP_KEY` will be auto-generated during setup
 > [!NOTE]
-> This is only required if another service is already running on the port `8001`
-- Set `APP_URL` to `http://localhost:<AVAILABLE_PORT>`
+> This is only required if the default port, `8001`, is not available.
+- Set `APP_URL` to `http://localhost:PORT
```

Is this what you had in mind?

edwh: Yup.

edwh:
@ardelato 

> @edwh could you share your WSL setup and steps you've taken? I can't seem to properly get the same results you are getting.

This feels like something which would work better if you ping me on Slack and we do a screen share.  

But:
* I start WSL on Windows, as my usual user.
* I do `sudo su -`
* `whoami` then shows me as `root`
* I have permissions to run `mix` , unlike you:
* 
image

edwh:
@ardelato I think this is still with you for some final doc tweaks?

docs/local-development.md
ardelato: Addressed in 3ce4fdd5f6d92fed39a229c152700ffc28cc7041

ardelato:
We should be good now. I included the changes for resolving the permission issues we were seeing in Windows and updated the docs accordingly.

ardelato:
@edwh I am not really sure what to do about the Docker errors `SonarCloud` is mentioning. This docker image is meant to only be used locally.

edwh:
> @edwh I am not really sure what to do about the Docker errors `SonarCloud` is mentioning. This docker image is meant to only be used locally.

I agree.  I've marked those two as safe.

PR #798, opened 2025-04-30
Revert back to Laravel Mix

ardelato:
## Description

This reverts 20d5726.

I started running into issues building and serving the assets with Vite.

Vite uses ES modules and treats them as first-class citizens. However, a lot of the packages we use have not had an update in quite a while so they are still using CommonJS syntax.

I have tried multiple solutions to address one problem after the next, but it became like the cartoon scene of trying to plug water leaks.

What I have not tried is to audit the corresponding packages to either find an alternative package or drop the package completely.
- `leaflet`
- `vue2-leaflet`
- `bootstrap-sortable.js` -- hardcoded file, but could probably be replaced with https://www.npmjs.com/package/sortablejs

For now, we'll revert back to using Mix until there is more time to fully flesh out the Vite upgrade.


ardelato:
Disregard ^

edwh:
I think this is unlikely to go well.  I've been planning to stick with the original build system until we've fleshed out the API enough to have a  pure SPA client using Nuxt.  At that point we can reuse the existing components (probably using Claude Code or similar to rework to Vue3 - I've done that with success elsewhere).  Then we can throw away the legacy build system.

PR #801, opened 2025-05-19
RES-2031 Prevent double submit of login form leading to 419 errors.

edwh:
There was some code which looked like it had the intention of doing this, but it didn't work.

We are relying on form submit rather than API calls to log in at the moment.  Take a simple approach of ignoring double submits within 1s.

### DIFF
--- a/resources/js/components/LoginPage.vue
+++ b/resources/js/components/LoginPage.vue
@@ -2,7 +2,7 @@
   <div>
     <div class="row row-expanded pb-3">
       <div class="col-lg-6 d-flex">
-        <form id="login-form" action="/login" method="post" class="card card__login col-12 panel">
+        <form id="login-form" action="/login" method="post" class="card card__login col-12 panel" ref="form" @submit.prevent="submit">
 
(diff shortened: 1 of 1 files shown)


resources/js/components/LoginPage.vue
```diff
@@ -117,13 +117,18 @@ export default {
     }
   },
   methods: {
-    login() {
+    submit() {
       // We've seen double submits of the login form, leading to 419 errors.  Prevent the user submitting twice by
-      // double-clicking.
+      // double-clicking by ignoring submits within 1 second of the last submit.
```
ngm, line 122: Any harm in making this time a bit longer?  e.g. 5 seconds.  

In my experience of having this happen when logging in via mobile, it's when browser autosubmit happens and you have to wait a while for the login to complete.  You press it again because you didn't realise the autocomplete also autosubmit.   If the login occurs in less than a second, you probably won't actually press the button again.

ngm (changes requested):

ngm (approved):

PR #835, opened 2026-04-16
feat: add living specifications system

edwh:
## Summary

- Introduces a **living documentation system** that embeds user stories directly in PHP controller code via PHP 8 attributes (`#[Feature]`, `#[UserStory]`, `#[NoStory]`)
- Adds `specs:extract` artisan command that uses nikic/php-parser AST analysis to extract **168 user stories** across **8 feature areas** and **6 personas** from 29 controllers into a JSON manifest
- Stories are further grouped **thematically** within each feature (e.g. Groups → Invitations, Manage volunteers; Events → Attendance & volunteers, Stats & data) so the spec pages mirror the "What X can do" structure
- Adds `@story:ControllerClass::method` docblock annotations to test files, linking **165 of 168 stories (98%)** to their tests
- Generates a browsable **VitePress static site** with dual navigation (by feature and by persona), platform overview on the landing page, AI-written narrative overviews, test coverage indicators, and GitHub source links
- Includes GitHub Actions workflow for automatic deployment to GitHub Pages on push to develop
- Removes the historical `features/` directory (100+ Gherkin files) — all relevant content has been folded into the new code-level annotations

**Live preview:** https://therestartproject.github.io/restarters.net/

## What this enables

- **Claude maintenance**: when modifying controller methods, the `#[UserStory]` attributes are visible inline, prompting updates when behaviour changes
- **CI validation**: `php artisan specs:extract --check` detects manifest drift, unannotated public methods, and stale narratives
- **Browsable spec site**: stakeholders can browse what the platform does without reading code, organised by feature area or user persona, grouped by theme within each feature
- **Test coverage tracking**: each story shows whether it has test coverage via `@story:` references in test files

## Feature areas

Events (53 stories), Groups (36), Users (25), Devices (14), Platform (20), Networks (15), Administration (3), Dashboard (2)

## Personas

Admin, Host, Restarter, NetworkCoordinator, Guest, ThirdParty

## Test plan

- [x] `php artisan specs:extract` generates valid manifest.json with 168 stories
- [x] 165/168 stories (98%) linked to test methods via `@story:` annotations
- [x] `npm run build` in specs-site/ produces working static site
- [x] All 8 feature pages and 6 persona pages render correctly
- [x] Stories grouped by theme within each feature on both feature and persona pages
- [x] Live preview deployed to GitHub Pages
- [ ] Verify existing PHPUnit tests still pass (attributes are inert metadata)
- [ ] Review a sample of controller annotations for accuracy


edwh:
Rebuilt this branch on current develop rather than rebasing — the original was ~600 commits behind and the @story docblock diffs no longer applied after the Vite migration and general drift.

What changed:
- Recreated the branch from current develop as 4 commits (system + docs, controller attributes, test annotations, one follow-up fixing a CLAUDE.md addition that got left out of the first pass).
- Carried over the standalone deliverables as-is: `app/Attributes/{Feature,NoStory,UserStory}.php`, `app/Console/Commands/SpecsExtract.php`, `docs/specs/manifest.json` + narratives, the standalone VitePress site in `specs-site/` (own `package.json`, doesn't touch the app's Vite/npm setup), the GitHub Pages workflow, and the design doc. Also re-applied the deletion of the legacy `features/` Gherkin directory — it hadn't been touched since this branch first diverged, so nothing since has been lost by removing it.
- Ported every `#[Feature]`/`#[UserStory]`/`#[NoStory]` attribute and every `@story:` test reference from the original branch's final state onto the current code, using a per-file 3-way text merge (merge-base version / current develop / branch tip). Where current develop had independently added PHP return-type hints on the same method signatures the branch was annotating, the merge preserves both.
- Verified nothing was dropped: counted attributes/references before and after — 168/168 `#[UserStory]`, 31/31 `#[Feature]`, 15/15 `#[NoStory]`, and 326/326 `@story:` test references all made it across untouched files or files that still exist with the same method names. No controller or test file referenced by the original branch had been deleted or renamed, so there was nothing to drop or note as stale.
- Validated with `php -l` on all 94 touched PHP files (via the running `restarters` container) and ran the VitePress `prebuild` script locally against the carried-over manifest — both succeed.

Needs human follow-up:
- `docs/specs/manifest.json` is carried over verbatim from the original branch's last generation and wasn't regenerated, since `php artisan specs:extract` needs a full app container this worktree doesn't have. Given the annotation counts match exactly, it should already be accurate, but it's worth running `php artisan specs:extract` once after merge (or as part of a follow-up) to confirm and refresh the `generatedAt` timestamp.
- Did not run the PHPUnit suite (shared test DB, not safe to run from a second checkout) or the specs:extract command itself — only static validation (lint + counts + the JS prebuild step) was possible here.

ngm:
@edwh I can't recall who needed to action this one?

edwh:
@ngm — this one is mine to action. Below is the review of the outstanding test-plan item ("review a sample of controller annotations for accuracy"). There are a few fixes to make and then `develop` to merge in (the branch is 68 commits behind). After that it needs a reviewer.

*(AI-assisted review by Claude, run at a maintainer's request. Each finding below was checked against the code.)*

**Sample:** 35 `#[UserStory]` attributes across events, groups, users, devices, networks, admin, exports, roles, skills and tags; about 15 `@story:` test links; all 15 `#[NoStory]` uses. Most stories are accurate and match the actual permission checks. The test links are solid: no mislinks found.

### Wrong or misleading stories

- **`ExportController` — all five stories** (`devicesEvent`, `devicesGroup`, `devices`, `groupEvents`, `networkEvents`) say Restarter or NetworkCoordinator. None of these methods checks permissions, and `routes/web.php` says outright that the export routes allow anonymous access because therestartproject.org/download-dataset calls them. The persona should be Guest or ThirdParty. As written, the spec suggests these exports are restricted when they're public.
- **`PartyController::create`** has a second story, "As a Host, I can create an online event without a physical location". `create()` only renders the blank form. The online flag is handled in `API\EventController::createEventv2`, which already has its own correct story, and `OnlineEventsTest` covers it. The extra story should go.

### Minor

- **`UserController` Repair Directory role story** is labelled Admin. The check is `repairdir_role()` (Repair Directory SuperAdmin/RegionalAdmin), not the platform Administrator role behind every other Admin story, so the label is ambiguous.
- **`CalendarEventsController::allEvents`** is `#[NoStory]`, but its four siblings, which are also iCal feeds protected by a secret token, all have stories. Either all of them are user-facing or none are.
- Some methods that Admin, Host and NetworkCoordinator can all use have a story for only one of those personas (e.g. `GroupController::edit`). That's understated rather than wrong.

### Checked and correct

- `API\DeviceController::createDevicev2` / `updateDevicev2` say "a Restarter can log a device at an event I attended". That's right: `userHasEditEventsDevicesPermission` allows confirmed attendees. The inline comment there ("Only hosts can add devices to events") is stale and misleading, and could be fixed while we're in there.

### Process gaps

- **The description overstates `--check`.** It says `--check` detects unannotated public methods. It doesn't: `checkManifest()` only compares a freshly generated manifest with the committed one. Nothing in CI runs it either. So new endpoints can land with no annotation and nothing flags them. `API\GroupController::listSummaryv2`, added on `develop` since this branch point, is already a case. Either add a real "every public controller method has `#[UserStory]` or `#[NoStory]`" check and wire it into CI, or narrow the description.
- **Latent mislink risk:** `SpecsExtract` indexes `@story:` references by unqualified `ShortClass::method`. `GroupController`, `NetworkController`, `DeviceController` and `UserController` each exist in both the web and `API\` namespaces. There's no collision today, but a same-named method in both would mislink silently.

### Non-annotation checks (fine)

- The attributes are inert DTOs, and the only consumer is the artisan command, which parses the code statically rather than using reflection at request time. So there's no runtime behaviour change.
- The Sonar workflow-permissions alerts are already fixed on the branch (job-level `permissions:`).
- Removing `features/` leaves no dangling references.

**To merge:** fix the export personas, drop the extra `create()` story, merge `develop` and annotate what came with it, and settle the `--check` wording or enforcement.

PR #862, opened 2026-05-27
RES-1995: rework groups page to have a map of groups (re-opened after revert)

edwh:
Re-opens replacement for #744, which was merged but then reverted on master (commit 78282c1c1d) because the Laravel 10 → 13 upgrade was bundled in unintentionally.

This PR contains the same code as the original #744 **plus fixes for the incomplete items that surfaced during review**:

- **Maps-of-groups feature** (NetworkPage tag CRUD, GroupMapAndList, etc.)
- **Laravel 10 → 13 upgrade**
- **PHP ^8.1 → ^8.3**
- **PHPUnit 9 → 10**
- **Migration** adding nullable `groups.image` column with back-fill

## Follow-up fixes added on this branch (TDD throughout)

- **`GroupsRequiringModeration`** was stubbed with a literal `TODO` string visible on the Groups page (commit da1b055552). Restored the `<GroupsTable :groupids="groupIds" approve />` call with a `groupIds` computed derived from the moderation store. 3 Jest tests.
- **`GroupsPage`** was receiving `showTags` / `networks` / `allGroupTags` from the blade but never forwarding them to its inner `GroupsTable`, so tag badges didn't render on the "Your groups" tab. Also fixed a `your-area="yourArea"` literal-string binding bug. Removed three genuinely-unused props (`yourLat`, `yourLng`, `userId`) from the component and the matching blade attributes. 2 Jest tests. (commit d9c127b043)
- **`groups/fetch`** action now de-dups concurrent in-flight requests for the same `(id, includeStats)` key — GroupsPage's `mounted()` loop dispatches one fetch per `yourGroup` in parallel, so without this the same group could be in flight twice. Module-scope Map with a `finally`-cleared entry so rejections don't poison the slot. 5 Jest tests. (commit d9c127b043)
- **Stale test comment** in `tests/Integration/grouptags.test.js` referring to chasing the Vue reactivity bug — fix shipped, comment updated. (commit d9c127b043)

A single empty commit (9beecf167f) was added at the start so GitHub would allow opening this PR from the branch (the original commits are still in develop's history via the merge ancestry, so GitHub otherwise reports 'no commits between').

## Database / migration note

Production already has the `groups.image` column from the original deploy of #744. The revert removed the migration *file* but the column and its `migrations` table row remain in prod. When this PR is re-merged the file comes back, Laravel sees the migration as already-ran (because the row is still there), skips it, and the back-filled data is preserved. No rollback needed.

## References
- Original PR: #744
- Revert commit on master: 78282c1c1d


edwh:
Superseded and closing. This PR bundled two independent changes — the *map of groups* feature and the Laravel 13 upgrade — which have been split into separate PRs against current `develop`:

- **#887** — map of groups (on Laravel 10)
- **#888** — Laravel 13 upgrade (framework only)

Both are green on CircleCI. Closing this combined PR to avoid confusion.

PR #872, opened 2026-06-10
Fix volunteer PII leak on public event pages

edwh:
## Summary

- **`Party::expandVolunteers()`** was serialising the full `User` Eloquent model into page HTML via `json_encode`. Any visitor to a public event page could read `api_token`, `calendar_hash`, `latitude`, `longitude`, `recovery`, and `mediawiki` tokens directly from the page source — and use the `api_token` to authenticate as that volunteer.
- **`User::$hidden`** only contained `password` and `remember_token`, leaving all other sensitive fields unprotected in any serialisation.
- **`ApiController::getUserInfo()`** was calling `makeHidden('api_token')` but leaving `calendar_hash`, `recovery`, etc. visible.

## Changes

1. **`app/User.php`** — extend `$hidden` to include `api_token`, `calendar_hash`, `recovery`, `recovery_expires`, `mediawiki`, `latitude`, `longitude`. Defence-in-depth: protects every serialisation point, not just the one we know about.
2. **`app/Party.php` `expandVolunteers()`** — replace the full `User` model assignment with an explicit whitelist of only the fields the frontend needs: `id`, `name`, optionally `email`, and the `user_skills` shape `{ skill_name: { skill_name: "…" } }`. The frontend only accesses `volunteer.name` and `volunteer.user_skills` — no other User fields are consumed.
3. **`app/Http/Controllers/ApiController.php`** — extend the belt-and-suspenders `makeHidden()` call to cover all credential fields.

## Tests

`tests/Feature/Events/VolunteerPiiTest.php` (3 tests, 26 assertions):
- `testEventPageDoesNotLeakVolunteerCredentials` — loads the public event page HTML with a known sentinel `api_token` and asserts it is absent, while the volunteer's name is present (sanity check that volunteers are still shown).
- `testUserModelHiddenFieldsExcludeCredentials` — asserts every sensitive field key is absent from `User::toArray()`.
- `testExpandVolunteersEmitsSafeShape` — asserts `expandVolunteers()` returns a plain array (not an Eloquent model) and contains none of the sensitive keys.

## Test plan
- [ ] All 3 `VolunteerPiiTest` tests pass (`OK (3 tests, 26 assertions)` confirmed locally)
- [ ] View a public event page in dev — attendee names and skills still visible
- [ ] View page source — no `api_token`, `calendar_hash`, coordinates visible


edwh:
Closing in favour of a clean branch rebased directly onto develop — the original branch was cut from a feature branch, so the diff included ~100 unrelated files. Replaced by #879

PR #880, opened 2026-06-19
Fix French category translations and pluralisation in lang.js

edwh:
## Fix French category translations and pluralisation

Category/cluster names render via `this.__(category.name)` — the English name is the top-level translation key (`lang/<locale>.json`, e.g. `"Desktop computer": "Ordinateur de bureau"`). Three things were needed to make this actually work and to get CI green:

### 1. Surface the JSON translations without crashing Vite
`laravelTranslations({ includeJson: true })` is **broken** in `vite-plugin-laravel-translations@0.3.1` — its JSON branch does `await import(`${cwd}/${absolutePath}`)` (doubled path, no JSON import attribute), which throws on Node 18+ and crashes Vite at config load → no assets → every request 500s (this was the "Wait for services" CI timeout, and why category names never translated). Replaced with a small wrapper plugin that runs the upstream plugin for the PHP translations and deep-merges `lang/<locale>.json` via plain `JSON.parse`. Verified via `vite resolveConfig`: `fr['Desktop computer'] = 'Ordinateur de bureau'`.

### 2. Pluralisation refactor + tests
`lang.js` now delegates to pure `lang-utils.js` (`translateWithLocale`/`choiceWithLocale`) with unit tests; `StatsImpact` pluralisation fixed to count the displayed items.

### 3. Translate the device name on the mobile repair podium
`GroupDeviceRepairPodium`'s desktop block used `{{ translatedName }}` but the mobile block rendered `{{ device.name }}` raw — fixed to use the translated value in both.

### CI
Also aborts Google Maps requests in the Playwright fixtures: Maps JS never finishes loading in headless Docker, so the page `load` event never fires and navigations hung past CircleCI's 10-min no-output timeout. With that, PHPUnit, Jest and Playwright all pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,14 +60,36 @@ jobs:
           no_output_timeout: 10m
 
       # Wait for core services to be ready (MySQL + web app). Discourse is started
-      # via docker:up-all but its readiness is checked separately (non-blocking)
-      # because it can take >10 min in CI while PHPUnit/Jest/Playwright tests do
-      # not require a live Discourse web process (PHPUnit mocks it).
+      # via docker:up-all but its readiness is checked separately (non-blocking).
+      # docker_run.sh runs npm install + npx playwright install + migrate:fresh --seed
+      # on every startup, which can take 20-25 min on cold CI machines. We poll
+      # directly here with a 25-minute ceiling rather than using the Taskfile helper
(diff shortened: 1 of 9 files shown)


ngm:
@edwh No change for me - still seeing category names in English when locale is French or French-Belgian.

PR #882, opened 2026-06-23
test: bind a deterministic geocoder globally to fix flaky "nearby" tests

edwh:
## Problem

`Dashboard\BasicTest::testUpcomingEvents` intermittently fails with:

```
ErrorException : Undefined array key "nearby"
tests/Feature/Dashboard/BasicTest.php:194
```

It passes in isolation and on other branches — classic flake. Root cause: the test depends on a **live** call to the Google Maps Geocoding API.

`App\Helpers\Geocoder::geocode()` does a `file_get_contents()` against `maps.googleapis.com` and returns `false` on any failure. That API is rate-limited and currently runs at ~20% errors (visible on the GCP dashboard). When the live geocode for the event's location fails, the event gets no coordinates, so it's excluded by the inner-joined `Party::scopeUpcomingEventsInUserArea` distance query, falls through to the "all upcoming" branch (which never sets `nearby`), and the assertion hits a missing key.

`BasicTest` was the one relevant test **not** mocking the geocoder — its siblings (`CreateEventTest`, `OnlineEventsTest`) each defined their own `GeocoderMock`, and three more files (`AttendanceTest`, `DeleteEventTest`, `AddRemoveVolunteerTest`) silently hit the live API too.

## Change

- Add a single deterministic `Tests\Support\GeocoderMock` and **bind it globally in `TestCase::setUp()`**, so no test ever calls the live API.
- Remove the duplicated per-test `GeocoderMock` + bindings from `CreateEventTest` and `OnlineEventsTest`.
- `GroupEditGeocodeFallbackTest` keeps its own failing mock (it exercises the geocode-failure path); its `setUp()` binding overrides the global one.
- The mock preserves the real `ForceGeocodeFailure` sentinel and resolves fixtures to central London (matching what the live API returned for the London-based test fixtures), so distance-based "nearby" assertions stay deterministic.

This also stops the test suite from contributing to live Geocoding API traffic.

## Verification

Ran locally (`task docker:test:phpunit`), all green:

| Test(s) | Result |
|---|---|
| `Dashboard/Groups/Style BasicTest` (`testUpcomingEvents`, `testPageLoads`) | OK (12 tests, 278 assertions) |
| `CreateEventTest` + `OnlineEventsTest` + `GroupEditGeocodeFallbackTest` | OK (20 tests, 271 assertions) |
| `AttendanceTest` | OK (2 tests, 22 assertions) |
| `DeleteEventTest` | OK (18 tests, 294 assertions) |
| `AddRemoveVolunteerTest` | OK (8 tests, 112 assertions) |

### DIFF
--- a/tests/Feature/Events/CreateEventTest.php
+++ b/tests/Feature/Events/CreateEventTest.php
@@ -7,7 +7,6 @@ use App\Events\UserLeftEvent;
 use App\EventsUsers;
 use App\Group;
 use App\Helpers\Fixometer;
-use App\Helpers\Geocoder;
 use App\Helpers\RepairNetworkService;
 use App\Listeners\AddUserToDiscourseThreadForEvent;
 use App\Listeners\RemoveUserFromDiscourseThreadForEvent;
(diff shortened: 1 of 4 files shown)


ngm (approved):
Looks good.  Good not to hit any geocoding API during testing.

PR #887, opened 2026-06-30
RES-1995: map of groups (Laravel 10)

edwh:
Reworks the groups page to show a **map of groups**. Split out from #862, which bundled this feature with a Laravel 10→13 upgrade — this PR is the **map feature alone, on the current Laravel 10 develop**. The L13 upgrade is a separate PR.

## What's here
- New Vue components: `GroupMap`, `GroupMarker`, `GroupMapAndList`, `GroupInfoModal`
- `GroupsPage` / `GroupsTable` / `GroupsRequiringModeration` reworked for the map + filter bar
- Groups now load client-side from a new `/api/v2/groups/summary` endpoint (+ `lat`/`lng` on the groups list, `GroupSummary` resource)
- `store/groups.js`: summary fetch + dedup
- Group filter-bar translation keys across locales; `leaflet-control-geocoder` JS dep
- Jest: `store` root + leaflet/vue-awesome mocks; new component + store tests
- `NetworkTest::network_page`: dropped the obsolete server-rendered `"networks":[id]` assertion (groups are now API-loaded)

## Stays on Laravel 10
No composer / framework / Docker / phpunit changes — verified `laravel/framework` is still `^10.48`.

## Verified locally
- Jest: 10 suites / 42 tests pass (all 5 map suites)
- Full PHPUnit suite: **518 tests, 0 failures**


edwh:
## Follow-up: bug fixes + network page map/list

A review of this PR against the RES-1995 discussion history turned up a set of regressions and the missing network-page half of the feature. All fixed here, test-first (each fix has a failing test committed with it).

### Summary

**Restored functionality**
- **Network page (`/networks/{id}`) now embeds the map + list of the network's groups** (was: a link + two lang keys that had been deleted from the lang files, so the page would have rendered literal `networks.show.groups_count` text). Starts zoomed out to frame all the network's groups; filters included.
- **`/group/network/{id}` filters by network again** — the `network` prop stopped at `GroupsPage` and never reached the map/list, so coordinators saw every group. `NetworkTest` had lost its assertion for this; a replacement guard asserts the network id reaches the Vue layer and the group is in the summary API data.
- **Filter bar is reachable again** (name / tag / location / country / network) on the map list and network views; the network dropdown is hidden on network-scoped views where it's redundant. The mobile "Show filters" toggle referenced a `toggleFilters` method that didn't exist — added.
- **"Groups requiring moderation" renders rows again** — `GroupsTable` reads from the `groups/list` store but moderation groups only existed in `groups/moderate`; they're now backfilled into the list store (without clobbering richer entries).

**Bug fixes**
- Your Groups tab showed "Follow group" for groups you already follow (`yourGroups` wasn't passed to the table); unfollowing now also flips the button immediately.
- Group logos in the list 404'd — the v2 APIs return a bare image path, the old server props were `asset('uploads/mid_…')`-prefixed. The table now prefixes bare paths.
- Clicking the Name column header threw a `TypeError` (sortCompare still read the old nested `group_name` shape); next-event sort compared the object rather than its start date.
- The list is now genuinely alphabetical — it sorted *after* slicing the page, so the first rows were "first N by id, alphabetised" rather than the alphabetically-first N.
- Panning to an area with no groups listed **all** groups (empty-in-view was treated the same as "map hasn't reported yet").
- `dragging` was enabled *only on mobile* (`!!L.Browser.mobile`) — desktop users couldn't pan the map. Also removed `gestureHandling: true`, which references a plugin that isn't installed.
- Green/red marker states referenced `marker-icon-green.png` / `marker-icon-red.png`, which don't exist in `public/images` — pins for followed groups rendered as broken images. Replaced with CSS `hue-rotate` classes on the stock blue marker.
- Groups without a geocode dropped markers at null island (0,0) — now skipped.
- `/api/v2/groups/summary` validated an `archived` parameter but ignored it (archived groups always returned) and lazy-loaded `networks`/`groupImage.image`/tags per group (O(N) queries — 6→12 queries just going from 3 to 9 groups in the test). Now honours `archived`, eager-loads, and returns `group_tags_full` so tag badges/filtering work from API data (`whenLoaded`, so other callers of `GroupSummary` don't pick up an N+1).
- Tab-switching no longer destroys and refetches the map (`lazy` → lazy-until-first-shown).
- Removed the inert `:interactive="false"` on `l-marker` (vue2-leaflet 2.7.1 ignores it — misleading), dead `expandGroups()` + unused imports, and other dead code.

### Code quality review
- All fixes were test-first; the two live-only bugs (missing `toggleFilters`, missing marker images) were caught by browser validation and then reproduced in Jest before fixing.
- `GroupSummary.group_tags_full` is gated on `whenLoaded` so `NetworkController`/moderation callers keep their current query profile.
- `GroupMap.allGroups` now uses the same guarded `(g.networks || []).some(...)` pattern as `effectiveGroupIds`.

### Verified
- Jest: **78 tests, 13 suites, all green** (was 42/10).
- PHPUnit: full suite green (see CI), including new `GroupSummaryApiTest` (archived param, tags, query-count-doesn't-scale).
- `npm run production` builds clean.
- Browser-validated against a seeded local instance: follow/unfollow states on both tabs, marker click → info modal (name/location/next event/Go to group), map persists across tab switches with no re-fetch, followed-group pin renders green, `/networks/{id}` shows the scoped map+list with filters and no raw lang keys, `/group/network/{id}` scopes correctly (0 groups for an empty network), mobile filter toggle, console clean.

### Future improvements / known edges
- Ungeocoded groups can't appear on the map, and once the map reports what's in view they drop out of the list below it (they still show in the pre-report fallback). If coordinators need to find ungeocoded groups, that wants a dedicated view.
- `/api/v2/groups/summary` is unauthenticated and includes unapproved groups — same data the old logged-in-only page embedded, and consistent with `/groups/names`, but flagging the widened exposure for an explicit decision.
- Individually-fetched groups (Your Groups tab before the summary loads) don't carry `group_tags_full`, so tag badges appear there only after the summary fetch. Pre-existing shape mismatch in the v2 single-group resource (`tags`).
- The `groups_title2_mobile` tab label is still "Nearest" while desktop says "Other groups".

### Test plan
- `npx jest` — 78 tests.
- `task docker:test:phpunit` — full suite, incl. `GroupSummaryApiTest`, `NetworkTest`, `BasicTest`, `GroupViewTest`.
- Manual: see "Verified" above.

edwh:
Heads-up on a cross-branch interaction with the translation-key removal in this PR.

I removed `groups.join_group_button_mobile` and `groups.leave_group_button_mobile` because `translations:check` reports them unused — which is correct **on develop**, where `GroupsTable.vue` was their only consumer and the map replaced it.

They are *not* unused on `nuxt-client` (#898). `client/app/components/groups/GroupJoinButton.vue` uses both:

```
{{ isMember ? t('groups.leave_group_button_mobile') : t('groups.join_group_button_mobile') }}
```

That branch's `CheckTranslations` scans `client/app/`, so it sees them as used and keeps them. Develop's copy has no `client/` to scan, so it reports them unused.

So when this lands on develop and `nuxt-client` next merges develop, `GroupJoinButton` will reference two keys that no longer exist — the mobile-width join/leave labels would come out blank. Worth re-adding them on the `nuxt-client` side as part of that merge (or dropping `GroupJoinButton`, which the group view page no longer uses — it renders `GroupActions` instead).

Flagging rather than changing anything here, since removing them is the right call for develop.

PR #902, opened 2026-09-02
Use a CARTO API key for map tiles

edwh:
CARTO now require an API key on their raster basemaps. Without one the tiles still load, but each is stamped **API KEY REQUIRED** — visible on every group, event and venue map on the site.

## Approach

The key is injected at runtime via `window.restarters`, next to the config already passed that way, rather than compiled into the bundle. `Dockerfile.fly` runs `npm run production` against a stub `.env` long before the deployed environment's secrets exist, so a `VITE_` variable would bake in an empty string and never see the Fly secret.

`leafletTiles()` appends the key when present and returns the bare URL when not, so an unconfigured environment degrades to the watermark rather than a broken map. The standalone `#event-map` in `app.js` had its own copy of the tile URL and now shares the helper.

The key is not secret in any meaningful sense — it is served to every visitor in the page — but keeping it in config keeps it out of the repo and lets it differ per environment.

## Secrets

Already staged as `CARTO_API_KEY` on `restarters`, `restarters-dev` and `restarters-yesterday` (staged, so it lands with this deploy — no extra restart). `FLY_PREVIEW_SECRETS` has been updated so previews get it too.

## Verification

- A tile fetched with the key returns clean; without it, watermarked. A bogus key returns a byte-identical watermarked tile, confirming the `?key=` parameter is read. Retina `@2x` works.
- 4 new unit tests for `leafletTiles()`; full Jest suite (29) passes; `npm run production` builds.
- Built bundle contains the runtime lookup and *not* the key.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -88,6 +88,9 @@ DISCOURSE_APIKEY=1234
 REPAIRDIRECTORY_URL=http://map.restarters.test
 
 MAPBOX_TOKEN=1234
+# API key for CARTO basemap tiles (https://carto.com/basemaps/apikey/).
+# Without one the map tiles are watermarked but still render.
+CARTO_API_KEY=
 GOOGLE_ANALYTICS_TRACKING_ID=UA-12345678-1
 GOOGLE_TAG_MANAGER_ID=GTM-1ABCDEF
 GOOGLE_API_CONSOLE_KEY=1234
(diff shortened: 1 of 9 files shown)


ngm (approved):
Looks good to me, and works on the preview.

PR #905, opened 2026-09-02
ci: verify production after deploy, and restart once if it is down

edwh:
## What happened today

`production` was pushed at **12:02:30**. `deploy-fly-prod` built the image and started a rolling update. The machine came up at **12:08:07** without passing its health check, and flyctl gave up:

```
Updating existing machines in 'restarters' with rolling strategy
Error: failed to update machine 7813112c002178: Unrecoverable error:
timeout reached waiting for health checks to pass for machine 7813112c002178
Exited with code exit status 1
```

The job exited 1 and stopped, leaving the machine `started` but out of the proxy pool. With `min_machines_running = 1` that is a total outage — every request got `no known healthy instances found for route tcp/443`. It stayed down **43 minutes**, until the machine was restarted by hand.

Restarting **that same image** brought it straight back. Nothing was wrong with the build. The deploy had everything it needed to recover and instead walked away.

## Change

Stop treating flyctl's exit code as the verdict, and ask the question that actually matters: *is the site serving?*

- Probe `https://restarters.net/robots.txt` from outside Fly — the same static path the Fly health check uses, so the answer doesn't depend on the database being reachable.
- If it isn't serving, restart each machine **once**, re-probing after each so we stop as soon as it recovers.
- Never loop. A restart that doesn't help means something a restart cannot fix, and that needs a person — so the job fails, dumps `flyctl status` and the last 50 log lines.
- A deploy that only passed because of the restart says so loudly. This path existing is not a reason to stop asking why a machine failed to come up healthy.

## Notes

- Verified: the YAML parses and both step scripts pass `bash -n`.
- `jq` isn't used anywhere else in this config, so its presence is no longer assumed — it's installed if missing.
- This is a mitigation, not a cure. The single machine is the reason an unhealthy instance means a full outage; a second machine would make a failed rolling update a non-event. Worth considering separately.
- Needs to reach the `production` branch to take effect, so it follows the usual develop → master → production route.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -243,10 +243,14 @@ jobs:
           command: |
             export FLY_API_TOKEN="FlyV1 $FLY_API_TOKEN"
 
-            # Probe from outside Fly, on the same path the Fly health check
-            # uses.  robots.txt is static, so a 200 means nginx is up and the
-            # proxy has a healthy instance to route to - it does not depend on
-            # the database being reachable.
+            # Two probes, because they answer different questions.
+            #
+            # robots.txt is static: a 200 means nginx is up and the proxy has an
+            # instance to route to.  That is also the path the Fly health check
+            # uses - which is exactly why it cannot be the only probe.  On
+            # 2026-09-03 restarters-dev served robots.txt, and reported a passing
+            # health check, while every database-backed page returned 500,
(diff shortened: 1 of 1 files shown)


ngm:
LGTM.

